PR #18, opened 2015-12-10
On OSX, add target library path to RPATH

staticfloat:
I had problems on OSX when installing to a non-standard `CMAKE_INSTALL_PREFIX` (Things like `corrade-rc` would crash because it couldn't load .`libCorradeUtility.0.dylib`)  This fixes the problems on my computer, but I'm sure it requires testing on other platforms.  To trigger the problem on your computer, try installing `corrade` and then `magnum` using `-DCMAKE_INSTALL_PREFIX=$(echo ~)/local`.

It's possible that this needs to be done to more projects than just `magnum` and `corrade`, but I have not tested.


mosra:
Fixed with 13e5f88f0a102e4bc11a3afd954035f8394325fc, see mosra/magnum#126 for detailed discussion.

PR #19, opened 2016-01-12
Fixes compilation on MSVC2015. interface is a reserved keyword 

jaynus:
In MSVC2015, 'interface' is a reserved keyword and causes compilation errors when referenced via the Corrade templates; the error is only present once the tempalte is actually instantiated. I've simply renamed its two instances to interfaceName for compilation to work.

### DIFF
--- a/src/Corrade/PluginManager/AbstractManager.cpp
+++ b/src/Corrade/PluginManager/AbstractManager.cpp
@@ -113,7 +113,7 @@ AbstractManager::AbstractManager(std::string pluginInterface, std::string plugin
     /* Find static plugins which have the same interface and have not
         assigned manager to them */
     for(auto p: _plugins.plugins) {
-        if(p.second->loadState != LoadState::Static || p.second->manager != nullptr || p.second->staticPlugin->interface != _pluginInterface)
+        if(p.second->loadState != LoadState::Static || p.second->manager != nullptr || p.second->staticPlugin->interfaceName != _pluginInterface)
             continue;
 
         /* Assign the plugin to this manager and initialize it */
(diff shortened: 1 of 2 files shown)


mosra:
Huh, interesting. Do you use any special compilation flags? Some very strict error level? Some special MS-specific C++ dialect? Because it works for me both locally and also [on the AppVeyor CI](https://ci.appveyor.com/project/mosra/corrade/branch/master) and I never heard that anyone encountered any problems with it, even on MSVC 2013 and even when compiling for Windows Store/Phone.

What is the exact error message? Could it be that it collides with some ugly `#define` from Windows headers?

jaynus:
I am adding/using /W4 /MP - W4 should only warn on it, and not actually error out. Besides that, it is strictly the CMake MSVC visual studio 2015 32-bit project defaults.

All of this is done from a bare-bones CMake project, MSVC2015 Update 1 with cmake.

The unchanged version produces:
Corrade/PluginManager/AbstractManager.h(220): error C2236: unexpected token 'struct'. Did you forget a ';'? 
Corrade/PluginManager/AbstractManager.h(220): error C2332: 'struct': missing tag name 

Intellisense produces: Error: Invalid order of type specifiers

mosra:
Ahh, right.

```
#define interface struct
```

AARGH. FUCK MACROS. FUCK WINAPI. Someone from stackoverflow says that the `objbase.h` header is the root of all evil. Are you including `<windows.h>` in your code? Try defining `WIN32_LEAN_AND_MEAN` before including that file, it should reduce amount of crap carried along really *a lot* (including that interface keyword).

Hopefully you understand that I really don't want to do random API changes just because someone from 1995 decided to be really "clever" and annoyed people for the next five decades. I had the same problem with `near`, `far` (defined also in `windows.h`), `None`, `Empty` (defined in `Xlib.h`) etc. If possible, I'd rather want to fix the problem and not the consequence, because I really don't want to maintain a blacklist of names that I can't use because of crappy platform-specific historical reasons.

jaynus:
I am not including it, no, but I am sure a library somewhere might be. It's a sneaky fucker that I tend to really try to avoid too so totally understand. I wasn't under the assumption that it was actually just a windows.h include problem; as MSVC is highlighting it as a real primitives, and not a definition inclusion. Which is odd.

If thats the case, I'll work around getting that retarded definition out of the way and it'll solve for me - I'll cancel this PR as I didn't look into the issue that much :)

PR #21, opened 2016-03-11
Support for colored output in Windows shell

mdietsch:
Things work a little different on Windows ;)

Sorry for the bad commit message but I github ignored the entered text when committing from online editor :(


mdietsch:
The related tests do currently fail. I am not sure how to handle these... :(

Is color information in stringstreams even reasonable on Windows?

src/Corrade/Utility/Debug.cpp
```diff
@@ -59,13 +108,16 @@ template<Debug::Color c> Debug::Modifier Debug::colorInternal() {
         if(!noSpaceBefore) debug << Debug::nospace;
         debug << code;
         if(noSpaceBefore) debug << Debug::nospace;
```
mosra: I think putting space at proper place before/after space is important also for the Windows implementation to avoid visual artifacts.

Similarly for the bold implementation below.

src/Corrade/Utility/Debug.cpp
```diff
@@ -120,11 +173,12 @@ void Debug::resetColor(Debug& debug) {
     if(debug._flags & InternalFlag::DisableColors) return;
 
     debug._flags &= ~InternalFlag::ColorWritten;
-
+#ifndef CORRADE_TARGET_WINDOWS
     const bool noSpaceBefore = !!(debug._flags & InternalFlag::NoSpaceBeforeNextValue);
     if(!noSpaceBefore) debug << Debug::nospace;
     debug << "\033[0m";
     if(noSpaceBefore) debug << Debug::nospace;
+#endif
```
mosra: This looks like resetting color back to default is not working on Windows at all. Or am I overlooking something?

src/Corrade/Utility/Debug.cpp
```diff
@@ -165,10 +219,11 @@ Error::Error(const Flags flags): Error{_globalErrorOutput, flags} {}
 
 Debug::~Debug() {
     if(_output && (_flags & InternalFlag::ValueWritten)) {
+#ifndef CORRADE_TARGET_WINDOWS
         /* Reset output color */
         if(_flags & InternalFlag::ColorWritten)
             *_output << "\033[0m";
-
+#endif
```
mosra: And also no color resetting implicitly on Debug destruction (end of a line, for example). I think these are crucial features to avoid leaking state (broken console colors for subsequent commands).

src/Corrade/Utility/Debug.cpp
```diff
@@ -42,6 +49,46 @@ template<> inline void toStream<Implementation::DebugOstreamFallback>(std::ostre
     value.apply(s);
 }
 
+#ifdef CORRADE_TARGET_WINDOWS
+HANDLE getStdHandle(const std::ostream* s) {
+    if(!s) return INVALID_HANDLE_VALUE;
```
mosra: This line is not needed I think.

src/Corrade/Utility/Debug.cpp
```diff
@@ -42,6 +49,46 @@ template<> inline void toStream<Implementation::DebugOstreamFallback>(std::ostre
...
+    switch(c) {
+        case Debug::Color::Black:   return 0;
+        case Debug::Color::Red:     return FOREGROUND_RED;
+        case Debug::Color::Green:   return FOREGROUND_GREEN;
+        case Debug::Color::Yellow:  return FOREGROUND_RED|FOREGROUND_GREEN;
+        case Debug::Color::Blue:    return FOREGROUND_BLUE;
+        case Debug::Color::Magenta: return FOREGROUND_RED|FOREGROUND_BLUE;
+        case Debug::Color::Cyan:    return FOREGROUND_GREEN|FOREGROUND_BLUE;
+        case Debug::Color::White:   return FOREGROUND_RED|FOREGROUND_GREEN|FOREGROUND_BLUE;
+        case Debug::Color::Default: return 0;
```
mosra: I think it would be better to have these values directly in the enum definition instead of doing a runtime switch. But I don't like having winapi includes in the public headers due to namespace pollution so it would need to be done without including the header (copypasting the corresponding `#define`s from the winapi header, for example).

mosra:
Hi, thank you a lot for your contribution!

I was reading about ANSI on Wikipedia and saw that DOS and pre-NT Windows supported ANSI escape codes so I thought this would be valid also for current Windows versions... I should have been reading more carefully, because

> The Win32 console did not support ANSI escape sequences at all until Windows 10 "Threshold 2".

In fact, I had tested it on [AppVeyor CI](appveyor.com) and it miraculously worked there (in contrast to the Win32 console) so I didn't expect any problem elsewhere :( However, it seems that AppVeyor doesn't and won't support the "proper" way (they say it's way too deep in Windows so it's impossible to capture the color info) so with your fix the colors are there no more :( It would be wonderful to have colors both on AppVeyor and normal Win32 console, but I guess that's kinda impossible.

> The related tests do currently fail. I am not sure how to handle these... :(
> 
> Is color information in stringstreams even reasonable on Windows?

The test was there just to verify that I'm putting the ANSI escape codes at proper places. From the look of it, I assume there is no way to do an unit test for the Windows implementation?

src/Corrade/Utility/Debug.cpp
mdietsch: You're right. The null-pointer case is implictitly handled. But sometimes I prefer expliciticity ;)

src/Corrade/Utility/Debug.cpp
mdietsch: This is not really a runtime switch. `c` is a template parameter and the compiler can optimize this switch easily to a single `return` statement.

I don't like copy&pasting from platform headers if not _really_ required.

src/Corrade/Utility/Debug.cpp
mdietsch: _Resetting_ is not implemented in Win32 API headers. I would have to store the previous state somewhere. I'll have a look into this.

src/Corrade/Utility/Debug.cpp
mdietsch: Same as above.

src/Corrade/Utility/Debug.cpp
mdietsch: You are just backing up restoring the state of  `InternalFlag::NoSpaceBeforeNextValue` which is getting reset in `debug << code`.
In the case of `CORRADE_TARGET_WINDOWS` this flag does not get touched and therefor does not need this backup/restore code.

mdietsch:
Yes, this whole Windows console color thing seems like a mess. :(
I also stumbled upon [ansicon] (https://github.com/adoxa/ansicon). But this whole DLL injection thing seems a bit too hacky for this very case here.

But nevertheless it seems that there are some consoles on Windows which support coloring with ANSI escape codes. So it would probably be better to not mutually exclude ANSI coloring and Win32.

One possible solution could be to always use Win32 API on windows **and** ANSI escape sequences dependant on the `DisableColors` flag. Then this flag should get renamed to `DisableAnsiColors` or something similar.

Or is there any other solution you would prefer?

src/Corrade/Utility/Debug.cpp
mosra: Yup, you're right, sorry.

mosra:
(Sorry, got interrupted and wasn't able to write everything at once.)

Hm... thinking about this, the coloring will work only when the program prints directly to the console without anything in betweeen. Which makes it all a lot less useful, because for example when I run unit tests via `ctest`, they get piped through some additional stream and thus there is no possibility to affect the console color anymore. And that was like my main use case for having colored output.

:(

If you can then please have a look into resetting the colors into the previous state and I'll merge this. I will make the ANSI/WINAPI switch a compile-time option later so the user can select between two options with different levels of sanity. Having a runtime switch or DLL injection seems too hacky, yeah.

Also, just a coding style nitpick, can you please indent the preprocessor `#ifdef`s the same as the surrounding code? Thank you.

mdietsch:
I did the requested changes.

But the more I think about this topic the more I think it would be a good idea to have both coloring methods (Win32 and ANSI codes) on Windows...

mosra:
Squashed into one commit and merged in adbc61707ab4ac425d88b06647304cae9d5b1f9b, thanks for the contribution! 

Switching between Win32 and ANSI colors is now possible with `UTILITY_USE_ANSI_COLORS` CMake option (defaults to `OFF`).

PR #25, opened 2016-07-15
Compiling with clang-cl

Squareys:
Hi @mosra , 

I did this quite a while ago, but it seems to still work. There were some fixes required in the CMakeFiles to distinguish clang-cl from MSVC:

```
For clang-cl, cmake sets CMAKE_CXX_COMPILER_ID to "Clang", the
CMAKE_CXX_COMPILER_VERSION to Clangs version (e.g. "3.9.0") and defines
MSVC.
```

To compile with clang-cl I use:

```
mkdir build
cd build
cmake .. -GNinja -DCMAKE_CXX_COMPILER=clang-cl -DCMAKE_C_COMPILER=clang-cl  -DCMAKE_INSTALL_PREFIX=...
cmake --build . --target install
```

(on CMake 3.6.0, in case that makes a difference).

Building with tests still results in an error. Also: I considered enabling clang-cl on ci, but had to find that it is neither available on chocolatey (for `cinst`) nor is there a .zip of the binaries provided on llvm.org. One would therefore have to compile llvm from source, which could take a while given AppVeyors speed.

I hope this helps as a start, though. It will at least let people compile corrade with the given commands.

Regards, Squareys.

### DIFF
--- a/src/Corrade/Containers/StaticArray.h
+++ b/src/Corrade/Containers/StaticArray.h
@@ -53,12 +53,12 @@ as more featureful alternative to plain C arrays or `std::array`.
 Usage example:
 @code
 // Create default-initialized array with 5 integers and set them to some value
-Containers::Array<int> a{5};
+Containers::StaticArray<5, int> a;
 int b = 0;
 for(auto& i: a) i = b++; // a = {0, 1, 2, 3, 4}
 
 // Create array from given values
-auto b = Containers::Array<int>::from(3, 18, -157, 0);
+Containers::StaticArray<4, int> b{3, 18, -157, 0};
 b[3] = 25; // b = {3, 18, -157, 25}
 @endcode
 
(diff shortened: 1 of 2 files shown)


Squareys:
Actually, tests *do* work and run successfully.

LB--:
I tried [`cmake -G "Visual Studio 14 2015 Win64" -T "LLVM-vs2014"`](https://gist.github.com/LB--/31f702b361e5dc46293435059b016417) and it seems to work, but I get some error messages within standard library headers - I don't think this is an issue with corrade, but I'm not sure how to fix it.

mosra:
Awesome! Will check that out.

Huh, now I found out that there is also Clang/C2 which is *bundled* with Visual Studio 2015 since Update 1. I wonder what are the differences and whether we could "just" support that on the CI and assume that clang-cl would work too.

([source](https://github.com/boostorg/hana/wiki/Setting-up-Clang-on-Windows#visual-studio-2015-with-clangc2) and the mentioned problem is [fixed in CMake 3.6](https://cmake.org/cmake/help/v3.6/release/3.6.html#generators))

Squareys:
@mosra @LB-- That is the `Visual Studio 14 2015` generator though. If you want to build with Ninja, `-T v140_clang_3_7` is not available.

Squareys:
Yeah, I also ran that from the VS bash, I forgot to mention that. It needs the environment variables defined with vcvarsall.bat.

PR #28, opened 2016-08-11
Interconnect: Use () instead of {} in constructor because of MSVC.

Squareys:
Hi @mosra !

**It's one of these again!**

This just popped up today. On MSVC I got a syntax error trying to compile `Corrade::Interconnect`.
A collegue reported ee4bbeb2e4c1003d7bbed4821ea074e8857f7b34 to compile fine (a few later may also work).

I found out that this is one of the cases where MSVC did not like the `{}` while calling a superconstructor.
Since the CI compiled without complaints, I am guessing this may be fine with ["Visual Studio Update 3"](https://www.visualstudio.com/en-us/news/releasenotes/vs2015-update3-vs)?
I will install that and report back whether that compiles without this change.

Greetings, Squareys.

### DIFF
--- a/src/Corrade/Interconnect/Emitter.h
+++ b/src/Corrade/Interconnect/Emitter.h
@@ -416,7 +416,7 @@ template<class Receiver, class ...Args> class MemberConnectionData: public BaseM
     public:
         typedef void(Receiver::*Slot)(Args...);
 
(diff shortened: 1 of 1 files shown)


Squareys:
Alright, so "Visual Studio Update 3" fixed the problem.

mosra:
Merged, thanks!

(Even though I'll update the min MSVC requirements to Update 3 later.)

PR #29, opened 2016-11-25
[WIP] MemorySanitizer and UndefinedSanitizer builds

mosra:
Putting these into a PR, because I lost all hope already.

- [ ] The MemorySanitizer needs libc++, which is *impossible* to get on Travis:
  - It is in the whitelist for Precise, but just adding `libc++-dev` to package list on Precise gets just ignored, because there's no such package the default repos and `apt-get` treats not found packages as regular expressions, which then matches `libc6-dev` or about anything else. So the installation succeeds, but silently doesn't install the package at all, failing during compilation with `can't find -lc++`. Whoever thought this is a good idea for a package manager?!!
  - It is not in the whitelist for Trusty.
  - The only PPA that has `libc++-dev` for Precise is `debian-sid`. Just enabling that PPA makes all hell break loose and apt-get install fails on some file conflict *much* later.

- [ ] The UndefinedSanitizer reports weird errors about mismatched class inheritance (false positive?) and apparently UBSan itself leaks, as it triggers a error from ASan on exit. See the build output for details.


mosra:
Totally by accident I found *some* info on the UBSan false positives related to vptr with dlopen(): https://whatofhow.wordpress.com/2015/03/17/odr-rtti-dso/

It references these commits, which look like they export polymorphic classes and load the dynamic objects using RTLD_GLOBAL:

- https://cgit.freedesktop.org/libreoffice/core/commit/?id=f0aa1a78fb209310e8baef53c02f365fca518d11
- https://cgit.freedesktop.org/libreoffice/core/commit/?id=b4f6b26b5a1a78fecfa95ec2eb7ac8b80495d8aa 

But I am already doing all of that? Another thing is using `-fvisibility=ms-compat` instead of `-fvisibility=hidden`, so maybe that's the key.

PR #30, opened 2017-01-09
doc: Fix (potentially misleading) typo

mdietsch:
No additional text


mosra:
Ow! Thank you!

*(Uh, proofreaded six times and this still slipped through...)*

PR #32, opened 2017-03-19
Fix compilation issue related to CORRADE_NORETURN being undefined

alicemargatroid:
CORRADE_NORETURN is defined in the Macros.h file, which is not included by Debug.h.
Adding it back restores compilability on modern versions of GCC/Clang

### DIFF
--- a/src/Corrade/Utility/Debug.h
+++ b/src/Corrade/Utility/Debug.h
@@ -34,6 +34,7 @@
 #include <tuple>
 #include <type_traits>
 
+#include "Corrade/Utility/Macros.h"
 #include "Corrade/Containers/EnumSet.h"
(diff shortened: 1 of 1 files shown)


alicemargatroid:
@mosra I doubt my addition of a header is causing IOS to break unit tests, so I hope this can be merged quickly.

mosra:
Thank you! 

Sorry for this. I see that this error pops out when compiling w/o deprecated features, which I somehow failed to test properly recently. I will add a special CI build just to cover the case of building w/o deprecated features, because I'm certain there are similar situations in other repositories as well.

Also I'm on fixing the iOS build -- there's some weird linker error and `xcpretty` failed in propagating the exit code so the build happily continued with running nonexistent unit tests :D

PR #33, opened 2017-04-01
Modernize Corrade

alicemargatroid:
This is a minor commit that meticulously modernizes Corrade with the following changes:

1. No moving elements that are better simply copied
2. push_back -> emplace_back when appropriate
3. Use defaulted constructor/destructors, not trivial ones
4. Use more move friendly constructors
5. Remove unnecessary quantifiers / add ones where appropriate
6. Range-for for all

Feel free to take what you like and leave what you don't!

### DIFF
--- a/src/Corrade/Utility/Arguments.cpp
+++ b/src/Corrade/Utility/Arguments.cpp
@@ -99,7 +99,7 @@ std::vector<std::string> Arguments::environment() {
     /* Standard Unix and local Emscripten environment */
     #if defined(CORRADE_TARGET_UNIX) || defined(CORRADE_TARGET_EMSCRIPTEN)
     for(char** e = environ; *e; ++e)
-        list.push_back(*e);
+        list.emplace_back(*e);
 
     /* System environment provided by Node.js. Hopefully nobody uses \b in
        environment variables. (Can't use \0 because Emscripten chokes on it.) */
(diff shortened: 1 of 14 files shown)


src/Corrade/Utility/ConfigurationValue.cpp
```diff
@@ -49,7 +49,7 @@ namespace Implementation {
     }
 
     template<class T> T BasicConfigurationValue<T>::fromString(const std::string& stringValue, ConfigurationValueFlags flags) {
-        std::string _stringValue = stringValue;
+        const std::string& _stringValue = stringValue;
```
mosra: Wow. Why did I ever do this? This speeds up parsing of all configuration files and command-line arguments by ∞ :D

src/Corrade/Utility/Arguments.cpp
```diff
@@ -141,7 +141,7 @@ std::vector<std::string> Arguments::environment() {
     return list;
 }
 
-Arguments::Arguments(std::string prefix): _prefix{prefix + '-'} {
+Arguments::Arguments(const std::string& prefix): _prefix{prefix + '-'} {
```
mosra: I did this because I thought the `operator+(string&&, char)` would avoid the copy, maybe, if there would be enough capacity in the original string (which it probably won't be, so this is a pessimization). And then I forgot the `std::move()` anyway. Twice. Argh.

mosra:
I rebased the changes on master and did one small change -- I don't like using `auto` if the type is not immediately clear from surrounding lines (and the type is not horrendously long to type), so I put concrete types into the range-for loops.

Thanks a lot for doing this! Very appreciated.

PR #34, opened 2017-06-10
[appveyor] Added vs2017 to the build images

sigman78:


mosra:
Oh wow, the VS2017 port is *this* simple? Thank you!

I'll try to reduce the amount of builds a bit, though -- MinGW build takes significant amount of time and it doesn't need to be run twice as it has no dependency on installed VS at all.

mosra:
Hmm -- there is something suspicious about MSVC 2017. Both 2015 and 2017 report themselves as "version 19" or "version 14", depending on where one looks:

2015 says:

```
-- The CXX compiler identification is MSVC 19.0.24215.1
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/amd64/cl.exe -- works
```

while 2017 says:

```
-- The CXX compiler identification is MSVC 19.10.25019.0
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2017/Community/VC/Tools/MSVC/14.10.25017/bin/Hostx64/x64/cl.exe -- works
```

The thing is that my CMake buildscripts are enabling some ugly workarounds for 2015 ("14") which I hope are resolved with 2017 ("15"), but now I have no way to distinguish between them, as there is no significant version difference. I need to investigate what's going on here.

sigman78:
Concerning this issue - I've tried my best to detect installed msvc version with _vswhere_ but the current state of the things and https://github.com/Microsoft/vswhere/issues/62 makes it's almost unusable with legacy Visual Studio 2015 image. However, the good news is: Visual Studio 2017 image contains both legacy and newer build tools and with the minor workarounds, it is possible to select correct _vcvarsall.bat_ and such. Maybe even parametrise from the appveyor.yml and add x86 builds alongside. 

More on this:
https://blogs.msdn.microsoft.com/vcblog/2017/03/06/finding-the-visual-c-compiler-tools-in-visual-studio-2017/
https://github.com/Microsoft/vswhere/
My hair pulling experiments:
https://github.com/sigman78/appveyor-2017

LB--:
> there is something suspicious about MSVC 2017

I think the IDE and compiler are versioned separately.

mosra:
After some investigation I found out that MSVC 2017 presents itself as version `19.10`, while 2015 is `19.0`. There is now a new variable `CORRADE_MSVC2017_COMPATIBILITY` that enables a subset of the 2015 workarounds for 2017. I hoped there won't be any of the workarounds needed, but they still are.

The AppVeyor config was merged with some modifications as 3c2ded33c14d07a2a73e3a8a21d02bc3c3639f67 -- hope that's okay. I'll now proceed with integrating your changes in the other projects. Thanks a lot for this!

PR #43, opened 2018-04-02
[WIP] Fix ArrayView bool conversion

mosra:
The problem, in short:

```cpp
int a[5];
Debug{} << bool(ArrayView<int>{a}); // prints true, as expected
Debug{} << bool(ArrayView<int>{a, 4}); // prints true
Debug{} << bool(ArrayView<int>{a, 0}); // prints true .. huh?

Debug{} << bool(ArrayView<int>{nullptr}); // prints false, as expected
Debug{} << bool(ArrayView<int>{nullptr, 0}); // prints false
Debug{} << bool(ArrayView<int>{nullptr, 4}); // prints false .. huh?
```

In other words, currently `bool` conversion is done from the data pointer, not from the size pointer. I think it should depend on the size rather than on the data, to make life easier when writing parsers and such. The following code will cycle forever, because `input` is always some non-null pointer, even though size eventually becomes zero and one has to pay extra attention and say `!input.empty()` or something:

```cpp
ArrayView<const char> input; // some larger string

while(input) { // forever :(
    std::size_t bytesRead = ...;
    input = input.suffix(bytesRead);
}
```

That's not really intuitive, as e.g. `std::istream` also returns `false` when reaching EOF.

- - - 

So, fixing the bool conversion like this (as partially done in the PR already) seems like a good idea (for me at least), but there's one problem -- the bool conversion operator is disabled on MSVC as it causes ambiguities for `operator+`, basically clashing with the pointer conversion operator, even though the bool conversion is explicit (huh?). So there are the following options:

- [ ] Leave `explicit operator bool()` like it currently is (`return _data;`) and update the documentation to say it basically doesn't do what would one expect. That makes it consistent with MSVC, but practically useless and potentially dangerous if accidentally used for the wrong purpose.
- [ ] FIx it, but since it's disabled on MSVC, have totally different behavior there, causing code to break in unexpected ways. Wrong solution.
- [ ] Investigate and try to fix the ambiguity on MSVC. Might be impossible or might be possible only on MSVC 2017, leaving 2015 with the same dangerous inconsistency as above.
- [ ] Add an explicit branch to the constructor which sets the data pointer to `nullptr` in case size is zero. That makes it consistent with Array behavior, but might break some code. (I know about at least one case in the OpenDDL parse where I'm treating zero-sized nullptr view and zero-sized non-nullptr view differently -- the first one indicating a parse error, the second indicating EOF.)

To ensure consistency, this applies also to the `StaticArrayView`, but there zero-sized views are problematic at the moment due to some non-ISO-standard zero-sized arrays being present in function signatures.

Opinions? Thank you :)

### DIFF
--- a/src/Corrade/Containers/Array.h
+++ b/src/Corrade/Containers/Array.h
@@ -328,7 +328,7 @@ class Array {
         /** @brief Whether the array is non-empty */
         /* Disabled on MSVC <= 2017 to avoid ambiguous operator+() when doing
            pointer arithmetic. */
-        explicit operator bool() const { return _data; }
+        explicit operator bool() const { return _size; }
         #endif
 
         /**
--- a/src/Corrade/Containers/ArrayView.h
+++ b/src/Corrade/Containers/ArrayView.h
@@ -148,7 +148,7 @@ template<class T> class ArrayView {
         /** @brief Whether the array is non-empty */
         /* Disabled on MSVC <= 2017 to avoid ambiguous operator+() when doing
            pointer arithmetic. */
-        explicit operator bool() const { return _data; }
+        explicit operator bool() const { return _size; }
         #endif
 
         /** @brief Conversion to array type */
@@ -497,7 +497,7 @@ template<std::size_t size_, class T> class StaticArrayView {
         /** @brief Whether the array is non-empty */
         /* Disabled on MSVC <= 2017 to avoid ambiguous operator+() when doing
            pointer arithmetic. */
-        explicit operator bool() const { return _data; }
+        constexpr explicit operator bool() const { return size_; }
         #endif
 
         /** @brief Conversion to array type */
--- a/src/Corrade/Containers/StaticArray.h
+++ b/src/Corrade/Containers/StaticArray.h
@@ -212,6 +212,13 @@ template<std::size_t size_, class T> class StaticArray {
         /** @brief Moving is not allowed */
         StaticArray<size_, T>& operator=(StaticArray<size_, T>&&) = delete;
 
+        #ifndef CORRADE_MSVC2017_COMPATIBILITY
+        /** @brief Whether the array is non-empty */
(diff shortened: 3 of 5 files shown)


mosra:
During a port to MSVC 2019 I discovered that the explicit bool conversion works properly with the `/permissive-` flag enabled. However it's not possible to detect presence of that flag and OTOH forcing the users to have this flag enabled might break 3rd party code that *expects* non-standard MSVC behavior.

They plan to make that flag enabled by default at some point (2038?) but until then we can only wait I'm afraid (and then five more years until we can drop support for compilers that don't have it enabled by default).

mosra:
Recently I'm thinking about removing the implicit pointer conversion, as it causes various problems when comparing containers in `TestSuite` (more often than not the common type of an Array and ArrayView is a *pointer*, which is wrong), and doesn't actually improve speed of element access in Debug builds (as it's either `operator*()` or `operator[]()` anyway). Plus with an `operator[]` we could do range-checking in debug builds (as the additional comparison overhead is [minimal compared to the function call](https://twitter.com/zeuxcg/status/1097882097975386112)).

OTOH, I am not sure if the above is actually a good thing to do. For example, `Directory::read()` could be updated to distinguish better between a "file not found" state and "file found but empty" state. Right now it returns `{nullptr, 0}` in both, but it could return `{nullptr, 0}` when the file is not found and `{somePointer, 0, customNoOpDeleter}` when the file is found but empty. Then, the following would work with the current state (but would break if the bool conversion change is done):

```cpp
if(auto file = Utility::Directory::read("foo.dat")) {
   // the file is found, but might be empty
}
```

This semantics makes much more sense, and the above parser snippet could be easily fixed to the following (while the above can't really be fixed unless taking the `file` out of the loop and testing for `file.data()`, which makes it much more verbose):

```cpp
ArrayView<const char> input; // some larger string

while(!input.empty()) {
    std::size_t bytesRead = ...;
    input = input.suffix(bytesRead);
}
```

PR #48, opened 2018-08-27
[WIP] Ability to use plugins written in Java with PluginManager

mosra:
Blergh, code from the 2012 version of me.

Pushed mainly for archival purposes, this was an Uni assignment and I managed to load plugins written in Java and use them from C++. Maybe this could be useful at some point (in Android?), so not throwing that away completely.


mosra:
The sole purpose of this PR was to archive the original code, closing to not make it look like there's something left to be done.

PR #55, opened 2019-01-21
Source location in Debug

mosra:
Similarly to the [new `dbg!` macro in Rust](https://blog.rust-lang.org/2019/01/17/Rust-1.32.0.html#the-dbg-macro), this implements source location for Debug using [std::experiemntal::source_location](https://en.cppreference.com/w/cpp/experimental/source_location). At the moment this feature is present only in GCC 8.1 / libstdc++ and up and a builtin in Clang 9:

- ~~no traces of `std::source_location` in libc++ yet as of November: https://github.com/llvm-mirror/libcxx/tree/master/include/experimental~~ using the builtins instead
- feature request for MSVC (it apparently won't be in C++20?): https://developercommunity.visualstudio.com/idea/354069/implement-c-library-fundamentals-ts-v2.html, tracking issue: https://github.com/microsoft/STL/issues/54

This is an experiment that needs further design iterations:

- [x] ~~At the moment, all `Debug{}`, `Error{}` etc. constructors are acquiring the source location and saving it even if it might not be used, which means *extreme* increase in binary sizes. While this is convenient and doesn't require changes on the user side (it's just enabling a flag and then all following debug statements in the same scope get source location), it's not efficient. Some ideas:~~
  - ~~Provide a dedicated function such as `debug()` that instantiates `Debug` and with source location (bad, it requires users to learn a new thing, also we would need to do an `error()`, `fatal()` etc. and there's a high chance these names will clash with user code)~~
  - ~~Provide a constructor tag such as `Debug{Debug::SourceLocation}` (again kinda bad, since it's verbose to say every time)~~
  - ~~Save source file name only at the point of enabling the debug output but save line in every constructor (would need some way to detect what file we are in in order to *not* print file info from unrelated files, maybe abusing unnamed namespaces?)~~
  - ~~Provide a constructor tag that explicitly does not use source location and then use that in all the library internals to minimize the impact (goes against "don't pay for what you don't use", doesn't really help users)~~
  - Use `!Debug{}` 
- [x] ~~The Rust `dbg!` macro works inside `if()` as well, provide something like that too?~~ not sure how
- [x] The GCC implementation requires C++14, ~~I'm working around that by `#undef constexpr` for C++11 (which is insane), submit a PR to libstdc++ that uses some C++14 constexpr instead?~~ -- using the builtins directly
- [x] ~~The file+line info output format is probably not the best (use colors? wrap in `[]` like Rust does? make it `format()`able, allowing stuff like column, date and other things there?)~~ postponed, will be done together with arbitrary line prefixing
- [x] Some way to print just file & line without anything else [like Rust 1.35 has](https://blog.rust-lang.org/2019/05/23/Rust-1.35.0.html#calling-dbg()-with-no-argument)

Obligatory docs:

image

Further design ideas welcome! :)

### DIFF
--- a/doc/corrade-changelog.dox
+++ b/doc/corrade-changelog.dox
@@ -45,6 +45,12 @@ namespace Corrade {
     existing @ref Containers::ArrayView<void> and
     @ref Containers::ArrayView<const void> types
 
+@subsubsection corrade-changelog-latest-new-utility Utility library
+
+-   Ability to optionally prefix @ref Utility::Debug output with a source file
+    and line on supported compilers. See @ref Utility-Debug-source-location for
+    more information.
+
 @subsection corrade-changelog-latest-changes Changes and improvements
 
 @subsubsection corrade-changelog-latest-changes-containers Containers library
--- a/doc/snippets/Utility.cpp
+++ b/doc/snippets/Utility.cpp
@@ -380,6 +380,19 @@ Utility::Debug{} << "this has default color again";
 }
 
 {
+/* [Debug-source-location] */
+float a = 336;
+
+!Utility::Debug{} << "the result is" << (a /= 8);
+!Utility::Debug{} << "but here it's" << (a /= 8);
+
+!Utility::Debug{};
+
+Utility::Debug{} << "and finally, " << (a *= 8);
+/* [Debug-source-location] */
+}
+
+{
 /* [Debug-nospace] */
 Utility::Debug{} << "Value:" << 16 << Utility::Debug::nospace << "," << 24;
 /* [Debug-nospace] */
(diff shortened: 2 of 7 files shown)


mosra:
Rust 1.35 has an [empty dbg!](https://blog.rust-lang.org/2019/05/23/Rust-1.35.0.html#calling-dbg()-with-no-argument) macro to print just the file & line. Added that to the list above.

mosra:
Invented a much better version that's opt-in, easy to use and doesn't bloat the binary when not used. The last remaining blocker is `#define constexpr`, but I might just ignore that problem for now.

mosra:
To have this working on both Clang and GCC, I'll probably change to using the compiler builtins, and that'll solve the `#define constexpr` as well.

mosra:
MSVC builtins seem to be implemented in upcoming 16.6: https://github.com/microsoft/STL/issues/54#issuecomment-586520516

PR #56, opened 2019-01-25
Add PKGBUILDs for MSYS2/MinGW-w64

williamjcm:
The one in the `corrade` subdirectory builds from the latest release of Corrade (2018.10 as of this writing), and *should* be suitable for submission to https://github.com/Alexpux/MINGW-packages.

Both were tested on my MSYS2 installation, using both i686 and x86_64 versions of MSYS2-provided MinGW-w64 (based on GCC 7.4.0 and 8.2.1 respectively).

If things need to be changed, don't hesitate to mention it.


package/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+pkgrel=1
+pkgdesc="C++11/C++14 multiplatform utility library"
+arch=('any')
+url="https://magnum.graphics/corrade/"
+license=('MIT')
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake" "${MINGW_PACKAGE_PREFIX}-ninja")
+options=(!strip)
+provides=('corrade-git')
+
+_rootdir=$startdir/../../
```
mosra: Can you put both files into `package/msys/` instead of directly in `package/`? Looks like this `_rootdir` variable is even expecting that :)

package/PKGBUILD
```diff
@@ -0,0 +1,34 @@
...
+# Contributor: mosra <mosra@centrum.cz>
+# Based on the packages/archlinux/PKGBUILD file by mosra <mosra@centrum.cz> and the MSYS2/MinGW PKGBUILD templates
+
+_realname=corrade
+pkgbase=mingw-w64-${_realname}-git
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}-git"
+pkgver=dev
+pkgrel=1
+pkgdesc="C++11/C++14 multiplatform utility library"
+arch=('any')
```
mosra: Interesting, for Arch this would mean the package has no native binaries and so can run on both 32bit and 64bit. But looking at the other MSYS packages, all of them seem to use `any` as well. That means the MSYS package contains both a 32bit and a 64bit build?

package/corrade/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+provides=('corrade')
+arch=('any')
+url='https://magnum.graphics/corrade/'
+license=('MIT')
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake" 'git' "${MINGW_PACKAGE_PREFIX}-ninja")
+source=("${_realname}"::"git+https://github.com/mosra/corrade.git#tag=v$pkgver")
+sha1sums=('SKIP')
+
+build() {
+  [[ -d "${srcdir}"/build-${CARCH} ]] && rm -rf "${srcdir}"/build-${CARCH}
```
mosra: Is removing the directory necessary? Personally I'm not doing that in any package and looking around, the MSYS [SDL2](https://github.com/Alexpux/MINGW-packages/blob/master/mingw-w64-SDL2/PKGBUILD) package doesn't do that either.

So unless that's required by some guidelines, I would remove this line :)

package/corrade/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+pkgver=2018.10
+pkgrel=1
+pkgdesc='C++11/C++14 multiplatform utility library'
+provides=('corrade')
+arch=('any')
+url='https://magnum.graphics/corrade/'
+license=('MIT')
+makedepends=("${MINGW_PACKAGE_PREFIX}-cmake" 'git' "${MINGW_PACKAGE_PREFIX}-ninja")
+source=("${_realname}"::"git+https://github.com/mosra/corrade.git#tag=v$pkgver")
+sha1sums=('SKIP')
```
mosra:
```sh
source=("https://github.com/mosra/corrade/archive/v${pkgver}.tar.gz")
sha1sums=('240c68bfe89a0cc86411b6854358b1607c06d90d')
```

Not sure why the original package was using git, that's much slower. Also no need for `git` in makedepends, then.

package/corrade/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+  else
+    extra_config+=("-DCMAKE_BUILD_TYPE=Debug")
+  fi
+
+  MSYS2_ARG_CONV_EXCL="-DCMAKE_INSTALL_PREFIX=" \
+    ${MINGW_PREFIX}/bin/cmake \
+      -G'Ninja' \
+      -DCMAKE_INSTALL_PREFIX=${MINGW_PREFIX} \
+      "${extra_config[@]}" \
+      ../${_realname}
```
mosra: After the change of `source` above, this needs to be `../${_realname}-${pkgver}`.

package/corrade/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+  #DESTDIR="$pkgdir" ninja -C build-${CARCH} install
+  cd "${srcdir}"/build-${CARCH}
+  DESTDIR="${pkgdir}" ninja install
+  
+  install -Dm644 "${srcdir}/${_realname}/COPYING" \
+    "${pkgdir}${MINGW_PREFIX}/share/licenses/${_realname}/COPYING"
+}
+
+# vim: ts=2 sw=2 et:
+# getver: no
```
mosra: I kept the two-space indentation in the original PKGBUILD because `@xyproto` uses that for all his other packages. But since we're not bound by that here, can you make it four-space-indented like the other file?

Also, no need for the two above lines anymore. The first is handled with a project-wide `.editorconfig`, the second was AFAIK needed only because of the git sources, which are no more.

package/corrade/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+
+  ninja
+}
+
+package() {
+  #DESTDIR="$pkgdir" ninja -C build-${CARCH} install
+  cd "${srcdir}"/build-${CARCH}
+  DESTDIR="${pkgdir}" ninja install
+  
+  install -Dm644 "${srcdir}/${_realname}/COPYING" \
```
mosra: ... and this needs to be `${srcdir}/${_realname}-${pkgver}/COPYING`.

package/corrade/PKGBUILD
```diff
@@ -0,0 +1,50 @@
...
+# Contributor: mosra <mosra@centrum.cz>
+# Based on the Corrade Arch Linux PKGBUILD maintained by xyproto and the MSYS2/MinGW PKGBUILD templates
+
+_realname=corrade
+pkgbase=mingw-w64-${_realname}
+pkgname="${MINGW_PACKAGE_PREFIX}-${_realname}"
+pkgver=2018.10
+pkgrel=1
+pkgdesc='C++11/C++14 multiplatform utility library'
+provides=('corrade')
```
mosra: :thinking: hmmm. Not sure about this line. Any reason why is it here? Otherwise I would just remove it.

williamjcm:
Sorry, I messed up the file upload. I'll try fix it to the best of my ability.

package/corrade/PKGBUILD
williamjcm: I copy-pasted it from the [templates](https://github.com/Alexpux/MINGW-packages/tree/master/mingw-w64-PKGBUILD-templates).

Maybe they have a reason for doing that.

package/corrade/PKGBUILD
williamjcm: Yeah, you're right. I'll remove it.

package/corrade/PKGBUILD
mosra: Some Arch templates had that too, but from my experience 90% packaging errors were from something else than broken build directory. So rather than waiting every time to get it built from scratch I would wipe the `build-*` directory manually in those remaining 10%.

This is usually needed with venerable Makefile-based systems where dependencies are not rebuilt correctly and such, not much of a problem here.

package/PKGBUILD
williamjcm: No.

My guess is that it allows both the i686 and x86_64 packages to be installed on both Win32 and Win64.

mosra: Okay. Let's keep this consistent with the remaining MSYS packages, then.

williamjcm:
Edits done.

mosra:
Pushed as cc49c1d9303448f3828389375f8e588fbcfdaa73, thanks a lot! :+1:

PR #58, opened 2019-02-23
Fix build on big-endian platforms

pkubaj:
There's no number variable, change it to value.

### DIFF
--- a/src/Corrade/Utility/Endianness.h
+++ b/src/Corrade/Utility/Endianness.h
@@ -65,7 +65,7 @@ class Endianness {
          */
         template<class T> static T bigEndian(T value) {
             #ifdef CORRADE_BIG_ENDIAN
-            return number;
+            return value;
             #else
             return swap(value);
             #endif


mosra:
Thank you! Sorry for such a silly mistake, big-endian platforms are hard to come by.

Since you have one, could I ask you a favor? :)  Can you build with tests as well (enable `BUILD_TESTS` in CMake), run them (using `ctest --output-on-failure` in the build dir) and report what failed? I bet there will be a few more bugs. Thanks a million! :+1:

pkubaj:
Sure.
```
The following tests FAILED:
         13 - ContainersStridedArrayViewTest (Failed)
         43 - UtilityDirectoryTest (Failed)
         47 - UtilitySha1Test (Failed)
         49 - UtilitySystemTest (Failed)
         53 - UtilityResourceTest (Failed)
```

The platform I use is Talos II from Raptor ( https://www.raptorcs.com/content/TL2B01/intro.html ), but if you want something cheaper, they offer preorders for Blackbird ( https://www.raptorcs.com/content/BK1B01/intro.html ) or you could just get PowerMac G5. Ubiquiti routers can also run big-endian and they are even cheaper (although also much slower).

mosra:
The `--output-on-failure` flag I mentioned should print also the output of the failing tests, that's kinda essential to knowing what's wrong and what should be fixed ;)

pkubaj:
Here's the complete log:
https://pastebin.com/GkSgVGyd

mosra:
Thanks a lot! Some of these are rather trivial, more about the test expectations being wrong, but some (such as the SHA-1 calculation or Resource) look actually serious. 

Would you mind if I bother you again later (next week, let's say) to test proposed fixes? :)

pkubaj:
No problem, I can test your fixes.

mosra:
Merged, thanks a lot! I opened #60 with the proposed fixes for big-endian test failures, would be great if you could have a look :)

PR #64, opened 2019-05-30
Fixing gentoo build instructions

mtao:
-99999 -> 9999 in the ebuild path


mosra:
Merged as 936c36301ca948ad5d9344f4a8db6fa642e81135. Thanks a lot!

PR #66, opened 2019-06-14
Utility/Debug: use thread_local for global state

xqms:
This allows using Debug/Warning/Error redirection in multi-threaded situations.

I roughly measured the performance impact on my system: 45ns (new) vs 37ns (old) for printing a single character (`Debug{} << "c"`), redirected into `std::stringstream`.
Benchmark code using https://github.com/david-grs/geiger is here: https://gist.github.com/xqms/d29764bcdeb0d0b49d5a4fd39497a4ec.

In my opinion the additional cost is worth the gain - especially since `Utility::Debug` is probably not used in perfomance-critical paths.

### DIFF
--- a/src/Corrade/Utility/Debug.cpp
+++ b/src/Corrade/Utility/Debug.cpp
@@ -78,13 +78,13 @@ HANDLE streamOutputHandle(const std::ostream* s) {
 
 }
 
-std::ostream* Debug::_globalOutput = &std::cout;
-std::ostream* Warning::_globalWarningOutput = &std::cerr;
-std::ostream* Error::_globalErrorOutput = &std::cerr;
+thread_local std::ostream* Debug::_globalOutput = &std::cout;
+thread_local std::ostream* Warning::_globalWarningOutput = &std::cerr;
+thread_local std::ostream* Error::_globalErrorOutput = &std::cerr;
 
 #if !defined(CORRADE_TARGET_WINDOWS) || defined(CORRADE_UTILITY_USE_ANSI_COLORS)
-Debug::Color Debug::_globalColor = Debug::Color::Default;
-bool Debug::_globalColorBold = false;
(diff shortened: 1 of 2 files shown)


xqms:
Urgh, it seems Xcode 7.3 does not support `thread_local`. I guess you want to keep supporting it?
I'll look for a workaround (probably some ugly map+mutex+this_thread::get_id() thing)...

mosra:
Hi, thanks, this is definitely something to have in, yes :)

There's `__thread_local` for Apple's Clang, it's used in [GL::Context as well](https://github.com/mosra/magnum/blob/8e1e52d6677e10d8af40a92a48affa241e274b1c/src/Magnum/GL/Context.cpp#L446-L452). Anything but `std::map` :D Thinking more about it, it probably makes sense to move the `MAGNUM_BUILD_MULTITHREADED` option into Corrade so people who *really don't want* the additional runtime or code size overhead (Emscripten apps might come to mind) can turn it off. I can do that during merge.

xqms:
Let's try this version with `__thread`. I don't have an Apple toolchain here, so I'm (ab)using the CI to check if things compile :-)

I kept the `#if __has_feature(cxx_thread_local)` since Apple supports `thread_local` since Xcode 8.0.

mosra:
Merged as cb0fe1326ffa698a2ea0a6516f48311cb14db392 and 2244c61d2c1dbb9e5fceb28daca00b24f4219f3f, thank you! On top I added the ability to opt-out (using `BUILD_MULTITHREADED`) for people that don't want the overhead (9749f55c1fcc2debc49638955718828988021f41).

I didn't have a chance to measure file size differences on Emscripten yet, in case the additional `thread_local` turns out to make a difference, I'll make it off by default there. Next step is also adapting Magnum to use this setting instead of its own independent one.

PR #71, opened 2019-06-26
Utility/Directory: fix documentation for path()

xqms:
The method returns everything before the *last* slash in the string.


mosra:
:man_facepalming: thanks! :laughing:

xqms:
One of the more fun PRs so far! You had me thinking for a moment, trying to find out in what weird context *that* functionality would be useful, before I looked into the source :laughing:
It's really a compliment to the quality of the API documentation that I trust it that much.

PR #74, opened 2019-08-14
UseCorrade.cmake: fixing .metadata's output directory

xiconxi:
UseCorrade.cmake: fixing custom-target<*.metadata> while`CMAKE_LIBRARY_OUTPUT_DIRECTORY` is setted

### DIFF
--- a/modules/UseCorrade.cmake
+++ b/modules/UseCorrade.cmake
@@ -520,14 +520,22 @@ function(corrade_add_plugin plugin_name debug_install_dirs release_install_dirs 
     endif()
 
     # Copy metadata next to the binary for testing purposes
-    add_custom_command(
-        OUTPUT ${plugin_name}.conf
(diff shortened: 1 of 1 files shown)


mosra:
Hi, thank you for the contribution!

I wonder, would this work also if `$<TARGET_FILE_DIR:${plugin_name}>` was used instead of `${CMAKE_CURRENT_BINARY_DIR}/${CMAKE_CFG_INTDIR}`? That would fix it to not require the extra branch.

The other function, `corrade_create_plugins_symlink_for_executable()`, is not used from any place and I think it could live in your codebase for now instead? As I mentioned in another issue, making this work better is on my list, so when that's done, this function shouldn't be needed anymore.

xiconxi:
I haven't use cmake-generator-expressions before, that's much better than mine。

mosra:
Commited in bb45552e65af119207c25ea760ca76ddafc3b568, hopefully it works for you the way it should :)

PR #77, opened 2019-09-15
modules: Fix test command when library output path is set

Squareys:
Hi @mosra !

This is a tiny one, but would help me with maintenance a bit if it were to be upstreamed :)

Best,
Jonathan

From the commit message:
~~~
E.g. when `LIBRARY_OUTPUT_PATH` is set globally or the target property
`LIBRARY_OUTPUT_DIRECTORY` is set.

add_test(<name> <command> <args>) does not resolve the target file, but
rather just runs the command, which works as long as it is where cmake
puts in by default, but not if the library output dir is modified.

add_test(NAME <name> COMMAND <command> <args>) on the other hand resolves
generator expressions and <command>, if an executable target, will
"automatically be replaced by the location of the executable created at
build time." (from the cmake docs.)
~~~

### DIFF
--- a/doc/corrade-changelog.dox
+++ b/doc/corrade-changelog.dox
@@ -48,6 +48,7 @@ namespace Corrade {
     @ref CORRADE_TARGET_DINKUMWARE macros to detect the common STL
     implementations
 -   New @ref CORRADE_BUILD_MULTITHREADED build-time option, moved from Magnum
+-   Fixed test execution command when library output path is set in CMake
 
 @subsubsection corrade-changelog-latest-new-containers Containers library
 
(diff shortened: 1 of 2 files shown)


mosra:
> Unlike the above NAME signature no transformation is performed on the command-line to support target names or generator expressions.

cmake **what the hell**.

Can you add a changelog entry as well so I don't have to? :)

Squareys:
Sure!

> cmake what the hell.

That's documented, though :D

mosra:
I mean, I see no reason at all why the shorthand version would need to behave *so differently*.

Squareys:
Done ✔️

mosra:
Merged as 929843482bc276b3b98219dae656f449a7b1795a, thanks!

Issue #81, opened 2019-11-13
vcpkg install x86-uwp error: Native `corrade-rc` executable not found


linuxaged:
maybe we need to make `corrade-rc` as a standalone dependency?

`$ vcpkg install corrade[*]  --triplet x86-uwp`

Error:

[...]

PR #82, opened 2019-11-15
Fix corrade-rc executable not found when build target WinRT #81

linuxaged:
building has passed on my machine.
but not sure what does "similar change would be in UseCorrade.cmake" mean in the TODO comment.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,7 +152,7 @@ endif()
 
 # Initialize macros etc.
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/modules/" ${CMAKE_MODULE_PATH})
-if(CMAKE_CROSSCOMPILING)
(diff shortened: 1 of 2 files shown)


mosra:
Thanks for working on this!

I just restarted the AppVeyor build as it failed due to some network error, unrelated to your change :)
For UseCorrade.cmake I just checked and apparently nothing is needed anymore (it might have been in the past, which is why the comment). I'll merge this together with a documentation update later today.

For the magnum side of things, regarding OpenAL .. this would need to have vcpkg packages updated with "per-platform defaults" as described by the vcpkg maintainers at https://github.com/mosra/magnum/pull/368. In the meantime installing magnum explicitly without audio should work (but yet again I forgot the correct command how to do that).

mosra:
Merged, thank you!

mosra:
Reverted in bb900a9c0150f8e3f9891649a8c5cb30533bed5c, since the feature apparently wasn't ever implemented in CMake.

PR #85, opened 2019-12-21
Utility: Minor improvements to Sha1 and AbstractHash

Squareys:
Hi @mosra !

As discussed via gitter, here is ArrayView input support for `Sha1 <<` and the removal of `DebugStl.h` from AbstractHash.

Cheers,
Jonathan

### DIFF
--- a/doc/snippets/Utility.cpp
+++ b/doc/snippets/Utility.cpp
@@ -33,6 +33,7 @@
 #include "Corrade/Utility/Arguments.h"
 #include "Corrade/Utility/Assert.h"
 #include "Corrade/Utility/Configuration.h"
+#include "Corrade/Utility/DebugStl.h"
 #include "Corrade/Utility/Directory.h"
 #if defined(CORRADE_TARGET_UNIX) || (defined(CORRADE_TARGET_WINDOWS) && !defined(CORRADE_TARGET_WINDOWS_RT)) || defined(CORRADE_TARGET_EMSCRIPTEN)
 #include "Corrade/Utility/FileWatcher.h"
@@ -40,6 +41,7 @@
 #include "Corrade/Utility/Format.h"
 #include "Corrade/Utility/FormatStl.h"
 #include "Corrade/Utility/Macros.h"
+#include "Corrade/Utility/Sha1.h"
 
 /* [Tweakable-disable-header] */
 #define CORRADE_TWEAKABLE
@@ -604,6 +606,25 @@ struct {
     float x, y;
 } position;
 };
+}
+
+{
+/* [Sha1-usage] */
+Utility::Sha1 sha1;
+
+/* Add 7 bytes of string data */
+sha1 << std::string{"corrade"};
+
+/* Add four bytes of binary data */
+const char data[] = { '\x35', '\xf6', '\x00', '\xab' };
+sha1 << Containers::arrayView(data);
+
+/* Print the digest as a hex string */
+Utility::Debug{} << sha1.digest().hexString();
+
+/* Shorthand variant */
+Utility::Debug{} << Utility::Sha1::digest("corrade");
+/* [Sha1-usage] */
 }
 }
 
--- a/src/Corrade/Utility/Sha1.cpp
+++ b/src/Corrade/Utility/Sha1.cpp
@@ -52,7 +52,7 @@ unsigned int leftrotate(unsigned int data, unsigned int shift) {
 
 Sha1::Sha1(): _dataSize(0), _digest{InitialDigest[0], InitialDigest[1], InitialDigest[2], InitialDigest[3], InitialDigest[4]} {}
 
-Sha1& Sha1::operator<<(const std::string& data) {
+Sha1& Sha1::operator<<(Containers::ArrayView<const char> data) {
     const std::size_t dataOffset = _buffer.empty() ? 0 : 64 - _buffer.size();
 
     /* Process leftovers */
@@ -65,7 +65,7 @@ Sha1& Sha1::operator<<(const std::string& data) {
         }
 
         /* Append few last bytes to have the buffer at 64 bytes */
-        _buffer.append(data.substr(0, dataOffset));
+        _buffer.append(data.prefix(dataOffset));
         processChunk(_buffer.data());
     }
 
@@ -73,10 +73,14 @@ Sha1& Sha1::operator<<(const std::string& data) {
         processChunk(data.data() + i);
 
     /* Save last unfinished 512-bit chunk of data */
-    _buffer = data.substr(dataOffset + ((data.size() - dataOffset)/64)*64);
+    _buffer = data.suffix(dataOffset + ((data.size() - dataOffset)/64)*64);
 
     _dataSize += data.size();
     return *this;
+}
+
+Sha1& Sha1::operator<<(const std::string& data) {
+    return *this << Containers::arrayView(data.data(), data.size());
 }
 
 /* GCC 6 (and possibly 7) on Raspberry Pi 3 Model B+ (aarch64) misoptimizes the
--- a/src/Corrade/Utility/Sha1.cpp
+++ b/src/Corrade/Utility/Sha1.cpp
@@ -26,6 +26,7 @@
 #include "Sha1.h"
 
 #include <cstddef>
+#include <string>
 
 #include "Corrade/Utility/Endianness.h"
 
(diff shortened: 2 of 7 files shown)


src/Corrade/Utility/Sha1.h
```diff
@@ -31,6 +31,7 @@
 
 #include <string>
 
+#include "Corrade/Containers/ArrayView.h"
```
mosra: Could be just the Containers.h header here and ArrayView.h included in the cpp.

src/Corrade/Utility/Sha1.cpp
```diff
@@ -65,20 +65,24 @@ Sha1& Sha1::operator<<(const std::string& data) {
         }
 
         /* Append few last bytes to have the buffer at 64 bytes */
-        _buffer.append(data.substr(0, dataOffset));
+        _buffer.append(data.prefix(dataOffset));
```
mosra: Nope, this won't work as soon as there is a 0 byte in the data. Same below. (Doesn't the test check this? It should!)

If you want to get fancy, the `_buffer` could be changed to a growable Array .. and there the append takes an arrayview, so this would work again.

Squareys: I guess that's a great opportunity to learn the growable Array API!

Squareys: It is not tested. I adapted `exact64bytes()` to test this case, computing the expected hash with the master version.

Squareys: Actually, this might require a specific use test case, since this will only fail if the `\0` is in the left-over part

mosra: yep, exactly .. nontrivial to test, but also nontrivial to discover when it gets broken, so a test is critical (thanks in advance!)

Squareys: Okay, so actually, this only happens in iterative use, if the previous data had lefovers and the new data contains a zero in the first `64-numLeftOverBytes` bytes.

src/Corrade/Utility/Sha1.h
Squareys: It's now needed for the

~~~
        /** @overload */
        template<std::size_t size>
        Sha1& operator<<(const char(&data)[size]) {
            return *this << Containers::arrayView(data, size);
        }
~~~

src/Corrade/Utility/Sha1.cpp
Squareys: @mosra So what I've discovered after moving the code to `char _buffer[128]` is that this substr() call up there ads a 0 delimiter to the end, which is not indended, correct?

The other substr call actually works for all the tests, because they end with 0 anyway (input is always a string).

I'll try to figure out how to get the intended hashes with a third-party application

Squareys: To generate the expected hashes, I added Sha1Test.py, it will print the expected hash per test case.

Squareys:
@mosra I addressed the feedback and moved the code to use `_buffer[128]` instead of `std::string`.

src/Corrade/Utility/Sha1.cpp
```diff
@@ -92,28 +99,34 @@ Sha1& Sha1::operator<<(const std::string& data) {
...
 #endif
 Sha1::Digest Sha1::digest() {
-    /* Add '1' bit to the leftovers, pad to (n*64)+56 bytes */
-    _buffer.append(1, '\x80');
-    _buffer.append((_buffer.size() > 56 ? 120 : 56) - _buffer.size(), 0);
+    /* Add '1' bit to the leftovers */
+    _buffer[_bufferSize++] = '\x80';
+
+    /* Pad to (n*64)+56 bytes */
+    const size_t padding = (_bufferSize > 56 ? 120 : 56) - _bufferSize;
```
mosra: `CORRADE_INTERNAL_ASSERT(_bufferSize + padding + 8 <= sizeof(_buffer))`? Since we don't have a string growing indefinitely anymore, it's important we don't overflow :)

src/Corrade/Utility/Test/Sha1Test.cpp
```diff
@@ -77,6 +81,22 @@ void Sha1Test::twoBlockPadding() {
...
+void Sha1Test::zeroInLeftover() {
+    Sha1 sha;
+    sha << std::string(
+        "123456789a123456789b123456789c123456789d123456789e123456789f12341\000134", 69);
+    sha << std::string("\0001", 2);
+    CORRADE_COMPARE(sha.digest(),
+        Sha1::Digest::fromHexString("5fdc3d8c862c3c3f86735c536824aee668f89967"));
+}
+
+void Sha1Test::noStringDelimiter() {
```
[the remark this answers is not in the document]
Squareys: This test I guess is superfluous now that I figured out that std::string::size() returns the byte size of the string without the delimiter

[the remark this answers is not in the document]
mosra: Ah :)

src/Corrade/Utility/Sha1.h
```diff
@@ -51,16 +52,26 @@ class CORRADE_UTILITY_EXPORT Sha1: public AbstractHash<20> {
...
         /** @brief Add data for digesting */
+        Sha1& operator<<(Containers::ArrayView<const char> data);
+
+        /** @overload */
         Sha1& operator<<(const std::string& data);
 
+        /** @overload */
+        template<std::size_t size>
+        Sha1& operator<<(const char(&data)[size]) {
+            return *this << Containers::arrayView(data, size);
```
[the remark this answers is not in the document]
Squareys: What if the const char isn't actually a string, but e.g. `char a[2]{1, 2}`? Imo, this should probably also include the delimiter (or maybe not exist at all?).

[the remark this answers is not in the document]
mosra: > not exist

but without this, it caused ambiguous overloads, no? Ideally it would be without this overload, yes. Any existing code that fails if you remove this one?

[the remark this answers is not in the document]
Squareys: In this case, probably the std::string override should also include the delimiter. That would diverge from what python doesn, though, I think.

Squareys:
@mosra Done. Thanks for all the feedback!

I also added a code snippet for the usage, I find those particularly useful when searching through the docs.

src/Corrade/Utility/Sha1.h
```diff
@@ -36,13 +36,22 @@
...
+@snippet Utility.cpp Sha1-usage
+
+*/
 class CORRADE_UTILITY_EXPORT Sha1: public AbstractHash<20> {
     public:
         /**
          * @brief Digest of given data
          *
          * Convenience function for @cpp (Utility::Sha1{} << data).digest() @ce.
+         * The @cpp '\0' @ce delimiter is not included.
```
mosra: I wouldn't mention this here at all, it's misleading -- what it does do when your string has a 0 byte in the middle? Or when `data.back() == '\0'`? The std::string historically didn't have to allocate extra space for the 0-delimiter (it does so only from c++11).

Same below and in the snippet above, it only adds confusion IMHO -- it takes all `size()` bytes from it, which is what one expects, so it doesn't need to be mentioned explicitly.

doc/snippets/Utility.cpp
```diff
@@ -605,6 +606,25 @@ struct {
...
+/* Add data with delimiter */
+sha1 << Containers::arrayView("edarroc", 8);
+
+/* Compute digest as a hex string */
+Utility::Sha1::Digest d = sha1.digest();
+std::string hexDigest = d.hexString();
+
+/* Compute digest of string without delimiter */
+Utility::Sha1::digest("corrade");
+/* [Sha1-usage] */
```
mosra: Err ... no, too confusing 😅 The below is not ideal either tho, I'd need to think more for a better example:

```cpp
Utility::Sha1 sha1;

/* Add 7 bytes of string data */
sha1 << std::string{"corrade"};

/* Add four bytes of binary data */
const char data[] = { '\x35', '\xf6', '\x00', '\xab' };
sha1 << Containers::arrayView(data);

/* Print the digest as a hex string */
Utility::Debug{} << sha1.digest().hexString();

/* Shorthand variant */
Utility::Debug{} << Utility::Sha1::digest("corrade");
```

src/Corrade/Utility/Sha1.h
Squareys: It cause a lot of confusion for me not knowing whether or not it does include the final delimiter and I originally expected it to, so having this somewhere would make sense to me... maybe as a note? If there is `\0` in the middle, it will already mess up there with the char* => string conversion and use strlen... maybe add a note for that? 🤔

Squareys: I removed this line.

doc/snippets/Utility.cpp
Squareys: I changed the snippet to this for now (+ DebugStl.h include).

Squareys:
@mosra I made those two changes (see answers to those comments).

src/Corrade/Utility/Sha1.h
mosra: > If there is \0 in the middle, it will already mess up there with the char* => string conversion and use strlen

Nope, you can do `std::string{"ab\0cde"}` and that'll make the string six characters long with a zero in the middle.

mosra:
Merged with a bunch of minor doc clarifications / commit squashing in 004f97474b2accc6516db6b6b7fb4644792d1a50...f6276772d3d513815f406c2133f1ad16874413db. Thank you!

PR #88, opened 2020-01-25
Utility::Debug: add std::ostream output operator call converter

rune-scape:
Compile errors are much clearer for types without a stream operator, and avoids being an actual catch-all.

This is a bit of a selfish pr, as my code to translate `Debug` `operator<<`s to `std::ostream` `operator<<`s is bugged when there is a catchall for Debug stream `operator<<`s.
Both stream fallbacks use each other. This results in being able to compile `cout << UnprintableClass{}`.
At runtime it gets stuck in recursion, flipflopping between both `operator<<`s. I thought that was pretty interesting.

Please tell me if I made any formatting mistakes.

### DIFF
--- a/src/Corrade/Utility/DebugStl.h
+++ b/src/Corrade/Utility/DebugStl.h
@@ -103,18 +103,36 @@ template<class ...Args> Debug& operator<<(Debug& debug, const std::tuple<Args...
 
 namespace Implementation {
 
+/* In order to ensure the most fitting operator<< overload gets called, we
+   can't simply check for *some* operator<< overload. As an example, for Array
+   there's both operator<<(ostream&, void*) and operator<<(Debug&, Iterable)
+   that can print it, but only the second one is desirable as the first prints
+   just a pointer. Directly checking for presence of either via
+
+    decltype(DeclareLvalueReference<std::ostream> << std::declval<T>()) and
+    decltype(DeclareLvalueReference<Debug> << std::declval<T>())
+
+   would yield "yes" in both cases, not telling us which one is better. Instead
+   we supply an object that's implicitly convertible to both as the first
+   argument, giving the overload resolution magic a chance to pick the better
+   fitting one. */
+struct OstreamOrDebug {
+    /*implicit*/ operator std::ostream&();
+    /*implicit*/ operator Debug&();
+};
+
 CORRADE_HAS_TYPE(
-    HasOstreamOperator,
+    HasBestFittingOstreamOperator,
     typename std::enable_if<std::is_same<
-        decltype(DeclareLvalueReference<std::ostream>() << std::declval<T>()),
+        decltype(std::declval<OstreamOrDebug>() << std::declval<T>()),
         std::add_lvalue_reference<std::ostream>::type
     >::value>::type
 );
 
 CORRADE_HAS_TYPE(
-    HasDebugStreamOperator,
+    HasBestFittingDebugOperator,
     typename std::enable_if<std::is_same<
-        decltype(DeclareLvalueReference<Debug>() << std::declval<T>()),
+        decltype(std::declval<OstreamOrDebug>() << std::declval<T>()),
         std::add_lvalue_reference<Debug>::type
     >::value>::type
 );
@@ -123,7 +141,7 @@ CORRADE_HAS_TYPE(
 struct DebugOstreamFallback {
     template<
         class T,
-        typename = typename std::enable_if<HasOstreamOperator<T>::value && !HasDebugStreamOperator<T>::value>::type
+        typename = typename std::enable_if<HasBestFittingOstreamOperator<T>::value>::type
     > /*implicit*/ DebugOstreamFallback(const T& t): applier(&DebugOstreamFallback::applyImpl<T>), value(&t) {}
 
     void apply(std::ostream& s) const {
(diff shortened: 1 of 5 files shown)


src/Corrade/Utility/DebugStl.h
```diff
@@ -103,9 +103,32 @@ template<class ...Args> Debug& operator<<(Debug& debug, const std::tuple<Args...
 
 namespace Implementation {
 
+template<typename T>
+typename std::add_lvalue_reference<T>::type declare_lvalue_reference() noexcept;
```
mosra: This is basically "like `std::declval<T>()`, but returning a T& instead of T&&", right? :) I think this might be useful in other contexts as well, will put it into `TypeTraits.h` as a public API.

src/Corrade/Utility/DebugStl.h
```diff
@@ -103,9 +103,32 @@ template<class ...Args> Debug& operator<<(Debug& debug, const std::tuple<Args...
...
+            decltype(declare_lvalue_reference<std::ostream>() << std::declval<C>()),
+            std::add_lvalue_reference<std::ostream>::type
+        >::type;
+
+        template<typename>
+        static constexpr std::false_type check(...);
+
+    public:
+        static constexpr bool value = decltype(check<T>(nullptr))::value;
+};
```
mosra: There's a (admittedly weirdly named) [CORRADE_HAS_TYPE(className, typeExpression)](https://doc.magnum.graphics/corrade/TypeTraits_8h.html#a19ce628bb598766db0fffb24c19fc9de) macro that already contains the above functionality, but your variant seems to be a lot simpler than what I stole from somewhere back in the day :) If I fix the macro to have the second parameter variadic (so the comma doesn't break it), I *think* we could then reduce the above to be just this (containing the potentially hard-to-understand template magic in a single place):

```cpp
CORRADE_HAS_TYPE(HasOstreamOperator, typename std::is_same<
    decltype(declareLvalueReference<std::ostream>() << std::declval<C>()),
    std::add_lvalue_reference<std::ostream>::type
>::type)
```

(Also adapting the naming to what Corrade uses for type traits.) I hope I didn't overlook something that would make this impossible tho.

mosra (commented):
The extremely huge compiler errors on a mismatched operator<< were a pain, so thank you for making everyone's life better :)

> This is a bit of a selfish pr

Why? :) When I saw the notification, I actually thought you'd be adding the inverse operator here -- making `Debug`-printable types work with `std::ostream`. I'm happy to accept that, you'd be definitely not the only one using it.

src/Corrade/Utility/DebugStl.h
mosra: `CORRADE_HAS_TYPE()` is fixed to allow this in 8d4eb5bc333d8b8e58e5bbbc9946999f17f1648d, and I think it would work with a minor change:

```cpp
CORRADE_HAS_TYPE(HasOstreamOperator, typename std::enable_if<std::is_same<
    decltype(declareLvalueReference<std::ostream>() << std::declval<C>()),
    std::add_lvalue_reference<std::ostream>::type
>::value>::type)
```

I assume you have some way to test the infinite recursion already -- can you confirm? :)

src/Corrade/Utility/DebugStl.h
bowling-allie: yes! `std::delcval` declares an rvalue refrence, so it picked up the `operator<<` for `Debug&&` instead of the ones wanted

src/Corrade/Utility/DebugStl.h
bowling-allie: yes! works perfectly. A very useful macro too

bowling-allie:
I figured I should separate this from the feature adding one, but if you want I can add it in.

mosra:
Adding the inverse operator would have a good side effect of testing that the above SFINAE restriction actually works and there's no infinite recursion anymore (while this alone doesn't check for that, and so I'm afraid of breaking it by accident).

So, by all means, add it in :)

bowling-allie:
The extra checks are to avoid ambiguous calls

src/Corrade/Utility/TypeTraits.h
```diff
@@ -163,4 +163,10 @@ template<class T> using IsStringLike = std::integral_constant<bool,
 
 }}
 
+/**
+@brief Like @cpp std::declval @ce, but declares an lvalue reference
+*/
+template<typename T>
+typename std::add_lvalue_reference<T>::type DeclareLvalueReference() noexcept;
```
mosra: Minor: I'm usually not putting a newline after `template<..>`, keeping the whole signature a single line. (Or generally not adding a newline unless really needed to keep the code vertically terse.)

src/Corrade/Utility/DebugStl.h
```diff
@@ -130,4 +149,14 @@ CORRADE_UTILITY_EXPORT Debug& operator<<(Debug& debug, Implementation::DebugOstr
 
 }}
 
+template<typename T>
```
mosra: One last thing: can you 

- add a short documentation block for this new function,
- mention this in the class documentation so people get to know it exists -- there's a "Printing STL types" section, so maybe extending that a bit ("Printing STL types, interaction with iostreams"?)
- and add a test in `Test/DebugTest.cpp`? There's `ostreamFallback` and `ostreamFallbackPriority`, so something similar (`ostreamDelegate`?) and also similarly checking the priority works for the inverse case, picking the ostream operator over the Debug one -- btw., don't forget to list the two new methods in the `addTests()` so these get properly called

Thank you! :+1:

mosra (commented):
This is great, the diff is so nicely small and tidy that at first I thought I was looking at the previous state with the operator<< not added yet :)

src/Corrade/Utility/DebugStl.h
bowling-allie: yes! wow `ostreamDelegate` is way better than any name I considered

bowling-allie:
a problem:
implicit conversions take priority over the operator<<(only because HasOstreamOperator only checks for any possible ostream print operator), meaning Container::Array get printed as a void *

a fix is making `HasOstreamOperatorExact` and `HasOstreamMemberOperatorExact`, and checking like 4 versions of each... but like.. no

mosra:
@bowling-allie finally today I got a totally random idea and it seems it works -- or at least made the test pass. Full details in the 3 new commits I just pushed to your branch (I hope that's okay, hehe). 

Can you check with your own code if everything is still behaving as expected and no circular recursion or anything happens again?

PR #92, opened 2020-03-14
Directory: Print an error when mkpath() returns false

Squareys:
Hi @mosra,

as discussed per gitter, here's the PR that adds some error messages to failures in `Directory::mkpath()`.

Cheers,
Jonathan

### DIFF
--- a/src/Corrade/Utility/CMakeLists.txt
+++ b/src/Corrade/Utility/CMakeLists.txt
@@ -104,6 +104,12 @@ if(WITH_UTILITY)
         list(APPEND CorradeUtility_PRIVATE_HEADERS Implementation/WindowsWeakSymbol.h)
     endif()
 
+    # Functionality specific to Windows builds
+    if(CORRADE_TARGET_WINDOWS)
+        list(APPEND CorradeUtility_SRCS Implementation/WindowsError.cpp)
+        list(APPEND CorradeUtility_PRIVATE_HEADERS Implementation/WindowsError.h)
+    endif()
+
     # Objects shared between main and test library
     add_library(CorradeUtilityObjects OBJECT
         ${CorradeUtility_SRCS}
@@ -194,8 +200,10 @@ if(NOT CMAKE_CROSSCOMPILING)
         Resource.cpp
         String.cpp)
     if(CORRADE_TARGET_WINDOWS)
-        # Needed for dealing with the design failure that's called "Unicode WINAPI"
-        list(APPEND CorradeUtilityRc_SRCS Unicode.cpp)
+        list(APPEND CorradeUtilityRc_SRCS
+            # Needed for dealing with the design failure that's called "Unicode WINAPI"
+            Unicode.cpp
+            Implementation/WindowsError.cpp)
 
         # Functionality specific to static Windows builds. Not really needed
         # for corrade-rc as it doesn't load any plugins yet but easier than
--- a/src/Corrade/Utility/Directory.cpp
+++ b/src/Corrade/Utility/Directory.cpp
@@ -81,6 +81,7 @@
 #endif
 #include <shlobj.h>
 #include <io.h>
+#include "Corrade/Utility/Implementation/WindowsError.h"
 #endif
 
 #include "Corrade/configure.h"
@@ -202,9 +203,6 @@ bool mkpath(const std::string& path) {
     if(path.back() == '/')
         return mkpath(path.substr(0, path.size()-1));
 
-    /* If the directory exists, done */
-    if(exists(path)) return true;
-
     /* If parent directory doesn't exist, create it */
     const std::string parentPath = Directory::path(path);
     if(!parentPath.empty() && !exists(parentPath) && !mkpath(parentPath)) return false;
@@ -214,11 +212,21 @@ bool mkpath(const std::string& path) {
     /* Unix, Emscripten */
     #if defined(CORRADE_TARGET_UNIX) || defined(CORRADE_TARGET_EMSCRIPTEN)
     const int ret = mkdir(path.data(), 0777);
-    return ret == 0;
+    if(ret != 0 && errno != EEXIST) {
+        Error{} << "Utility::Directory::mkpath(): error creating" << path << Debug::nospace << ":" << strerror(errno);
+        return false;
+    }
+    return true;
 
     /* Windows (not Store/Phone) */
     #elif defined(CORRADE_TARGET_WINDOWS) && !defined(CORRADE_TARGET_WINDOWS_RT)
-    return CreateDirectoryW(widen(path).data(), nullptr) != 0 || GetLastError() == ERROR_ALREADY_EXISTS;
+    if(CreateDirectoryW(widen(path).data(), nullptr) == 0 && GetLastError() != ERROR_ALREADY_EXISTS) {
+        Error{} << "Utility::Directory::mkpath(): error creating"
+            << path << Debug::nospace << ":"
+            << Utility::Implementation::windowsErrorString(GetLastError());
+        return false;
+    }
+    return true;
 
     /* Not implemented elsewhere */
     #else
--- a/src/Corrade/Utility/Directory.cpp
+++ b/src/Corrade/Utility/Directory.cpp
@@ -437,8 +437,8 @@ std::string libraryLocation(const void* address) {
     #elif defined(CORRADE_TARGET_WINDOWS) && !defined(CORRADE_TARGET_WINDOWS_RT)
     HMODULE module{};
     if(!GetModuleHandleExA(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS|GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT, reinterpret_cast<const char*>(address), &module)) {
-        /** @todo use FormatMessage() for a string error */
-        Error{} << "Utility::Directory::libraryLocation(): can't get library location:" << GetLastError();
+        Error{} << "Utility::Directory::libraryLocation(): can't get library location:"
+            << Utility::Implementation::windowsErrorString(GetLastError());
         return {};
     }
 
--- a/src/Corrade/Utility/Implementation/WindowsError.cpp
+++ b/src/Corrade/Utility/Implementation/WindowsError.cpp
@@ -1,0 +1,56 @@
+/*
+    This file is part of Corrade.
+
+    Copyright © 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016,
+                2017, 2018, 2019 Vladimír Vondruš <mosra@centrum.cz>
+    Copyright © 2020 Jonathan Hale <squareys@googlemail.com>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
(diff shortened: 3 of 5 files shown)


src/Corrade/Utility/Directory.cpp
```diff
@@ -202,23 +202,39 @@ bool mkpath(const std::string& path) {
...
+        return false;
+    }
+    return true;
 
     /* Windows (not Store/Phone) */
     #elif defined(CORRADE_TARGET_WINDOWS) && !defined(CORRADE_TARGET_WINDOWS_RT)
-    return CreateDirectoryW(widen(path).data(), nullptr) != 0 || GetLastError() == ERROR_ALREADY_EXISTS;
+    if(CreateDirectoryW(widen(path).data(), nullptr) == 0 && GetLastError() != ERROR_ALREADY_EXISTS) {
+        WCHAR* errorStringW = nullptr;
+        FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM|FORMAT_MESSAGE_ALLOCATE_BUFFER,
```
mosra:
```suggestion
        FormatMessageW(FORMAT_MESSAGE_FROM_SYSTEM|FORMAT_MESSAGE_ALLOCATE_BUFFER,
```

Rule: always explicitly use the `W` APIs, to avoid non-Unicode-ready accidents :)

src/Corrade/Utility/Directory.cpp
```diff
@@ -202,23 +202,39 @@ bool mkpath(const std::string& path) {
...
+            nullptr, GetLastError(), 0, reinterpret_cast<LPWSTR>(&errorStringW),
+            0, nullptr);
+        Containers::ScopeGuard e{errorStringW, LocalFree};
+        std::string errorStringUtf8 = Unicode::narrow(errorStringW);
+        /* The error message already contains a newline */
```
mosra: There's a flag to disable the newline, I'd rather see that than disabling the newline on our side :)

src/Corrade/Utility/Directory.cpp
```diff
@@ -202,23 +202,39 @@ bool mkpath(const std::string& path) {
...
-    /* If the directory exists, done */
-    if(exists(path)) return true;
-
     /* If parent directory doesn't exist, create it */
     const std::string parentPath = Directory::path(path);
-    if(!parentPath.empty() && !exists(parentPath) && !mkpath(parentPath)) return false;
+    if(!parentPath.empty() && !exists(parentPath) && !mkpath(parentPath)) {
+        Error{} << "Utility::Directory::mkpath(): error: could not create parent path";
+        return false;
+    }
```
mosra: Assuming `mkpath(parentPath)` already prints an error message, I think this message is superfluous :)

src/Corrade/Utility/Directory.cpp
```diff
@@ -202,23 +202,39 @@ bool mkpath(const std::string& path) {
...
+    if(CreateDirectoryW(widen(path).data(), nullptr) == 0 && GetLastError() != ERROR_ALREADY_EXISTS) {
+        WCHAR* errorStringW = nullptr;
+        FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM|FORMAT_MESSAGE_ALLOCATE_BUFFER,
+            nullptr, GetLastError(), 0, reinterpret_cast<LPWSTR>(&errorStringW),
+            0, nullptr);
+        Containers::ScopeGuard e{errorStringW, LocalFree};
+        std::string errorStringUtf8 = Unicode::narrow(errorStringW);
+        /* The error message already contains a newline */
+        Error{Debug::Flag::NoNewlineAtTheEnd} <<
+            "Utility::Directory::mkpath(): error:" << errorStringUtf8.data();
```
mosra: why the `.data()`? :) Not a good practice, a null byte somewhere inside the error message (for example due to random memory passed as filename) would cut the message off.

src/Corrade/Utility/Test/DirectoryTest.cpp
```diff
@@ -585,12 +587,35 @@ void DirectoryTest::mkpathNoPermission() {
...
+    /* Ensure errors are printed in English */
+    std::setlocale(LC_ALL, "en_US.utf8");
+
+    std::ostringstream out;
+    Error err{&out};
+
     CORRADE_VERIFY(!Directory::mkpath("/nope/never"));
+
+    CORRADE_COMPARE(out.str(),
+        "Utility::Directory::mkpath(): error: Permission denied\n"
```
mosra: seeing this, i think it could make sense to print the actual path that failed -- which might be in 90% cases *the* most useful piece of information, the reason why it failed being secondary:

```suggestion
        "Utility::Directory::mkpath(): error creating /nope: Permission denied\n"
```

Same for the Windows case of course :)

src/Corrade/Utility/Test/DirectoryTest.cpp
```diff
@@ -585,12 +587,35 @@ void DirectoryTest::mkpathNoPermission() {
     if(Directory::exists("/nope"))
         CORRADE_SKIP("Can't test because the destination might be writeable");
 
+    /* Ensure errors are printed in English */
+    std::setlocale(LC_ALL, "en_US.utf8");
```
mosra:
```suggestion
    /* Ensure errors are printed in English */
    const char* currentLocale = std::setlocale(LC_ALL, nullptr);
    std::setlocale(LC_ALL, "en_US.utf8");
    ScopeGuard restoreLocale{currentLocale, [](const char* locale) {
        std::setlocale(LC_ALL, locale);
    }};
```

... and put it together with the `mkpath()` in a scope that's as small as possible -- I don't know what all `setlocale()` affects, so better be safe than waste time debugging weird interactions on weird platforms later :)

src/Corrade/Utility/Directory.cpp
```diff
@@ -429,8 +445,15 @@ std::string libraryLocation(const void* address) {
...
-        Error{} << "Utility::Directory::libraryLocation(): can't get library location:" << GetLastError();
+        WCHAR* errorStringW = nullptr;
+        FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM|FORMAT_MESSAGE_ALLOCATE_BUFFER,
+            nullptr, GetLastError(), 0, reinterpret_cast<LPWSTR>(&errorStringW),
+            0, nullptr);
+        Containers::ScopeGuard e{errorStringW, LocalFree};
+        std::string errorStringUtf8 = Unicode::narrow(errorStringW);
+        /* The error message already contains a newline */
+        Error{Debug::Flag::NoNewlineAtTheEnd} <<
+            "Utility::Directory::libraryLocation(): can't get library location:" << errorStringUtf8.data();
```
mosra: Hmm... maybe put this in a helper function in an unnamed namespace so it's not duplicated, or the `Implementation/WindowsError.h` we talked about?

src/Corrade/Utility/Directory.cpp
Squareys: Oh, I guess DebugStl is already included then 🤔

mosra: Yep, this code is full of strings anyway so not including it wouldn't save anything.

src/Corrade/Utility/Directory.cpp
Squareys: ✅

src/Corrade/Utility/Directory.cpp
Squareys: Are you sure? I found a format escape sequence, but that doesn't work with `FORMAT_MESSAGE_FROM_SYSTEM`, since it doesn't use the format string 🤔

src/Corrade/Utility/Directory.cpp
Squareys: ✅

src/Corrade/Utility/Directory.cpp
Squareys: ✅

src/Corrade/Utility/Test/DirectoryTest.cpp
Squareys: ✅

src/Corrade/Utility/Test/DirectoryTest.cpp
Squareys: ✅

src/Corrade/Utility/Directory.cpp
mosra: Eh FFS, what an API :roll_eyes: Sorry, missed that. 

Alternative solution -- cut the last character off :)

src/Corrade/Utility/Directory.cpp
Squareys: ✅ (needs some cleanup, though, WIN32_LEAN_AND_MEAN etc)

src/Corrade/Utility/Directory.cpp
Squareys: Sure 👍

src/Corrade/Utility/CMakeLists.txt
```diff
@@ -197,6 +203,8 @@ if(NOT CMAKE_CROSSCOMPILING)
         # Needed for dealing with the design failure that's called "Unicode WINAPI"
         list(APPEND CorradeUtilityRc_SRCS Unicode.cpp)
 
+        list(APPEND CorradeUtilityRc_SRCS Implementation/WindowsError.cpp)
```
mosra:
```suggestion
        list(APPEND CorradeUtilityRc_SRCS 
            # Needed for dealing with the design failure that's called "Unicode WINAPI"
            Unicode.cpp 
            Implementation/WindowsError.cpp)
```

src/Corrade/Utility/Directory.cpp
mosra: Yeah so this change could be completely reverted ;)

src/Corrade/Utility/Implementation/WindowsError.cpp
```diff
@@ -0,0 +1,49 @@
...
+#include "Corrade/Utility/Unicode.h"
+#include "Corrade/Containers/ScopeGuard.h"
+
+#define WIN32_LEAN_AND_MEAN 1
+#define VC_EXTRALEAN
+#include <windows.h>
+
+namespace Corrade { namespace Utility { namespace Implementation {
+
+std::string windowsErrorString(uint32_t errorCode) {
```
mosra:
```suggestion
std::string windowsErrorString(unsigned int errorCode) {
```

to be consistent with the definition in the header

src/Corrade/Utility/Implementation/WindowsError.cpp
```diff
@@ -0,0 +1,49 @@
...
+    WCHAR* errorStringW = nullptr;
+    FormatMessageW(FORMAT_MESSAGE_FROM_SYSTEM|FORMAT_MESSAGE_ALLOCATE_BUFFER,
+        nullptr, errorCode, 0, reinterpret_cast<LPWSTR>(&errorStringW),
+        0, nullptr);
+    Containers::ScopeGuard e{errorStringW, LocalFree};
+    /* Convert to UTF-8 and cut off final newline that FormatMessages adds */
+    return Unicode::narrow(Containers::arrayView<const wchar_t>(
+        errorStringW, wcslen(errorStringW) - 1));
```
mosra:
```suggestion
    return Unicode::narrow(Containers::arrayView(errorStringW, wcslen(errorStringW)).except(1));
```

is safer (would assert in case the error string is empty), also the explicit template shouldn't be needed

src/Corrade/Utility/Test/DirectoryTest.cpp
```diff
@@ -585,12 +587,37 @@ void DirectoryTest::mkpathNoPermission() {
...
+    Error err{&out};
+
     CORRADE_VERIFY(!Directory::mkpath("W:/nope"));
+
+    /* On Windows we cannot enusre messages are printed in a certain
+     * language, as they depend on the user's installed languages */
+    const std::string output = out.str();
+    const size_t newlinePos = output.find('\n');
+    CORRADE_VERIFY(newlinePos != std::string::npos);
+    CORRADE_COMPARE(output.substr(0, 49), "Utility::Directory::mkpath(): error creating W:: ");
```
mosra: Not sure if the first two lines are still relevant... maybe instead do

```suggestion
    CORRADE_COMPARE(output.substr(0, 49), "Utility::Directory::mkpath(): error creating W:: ");
    /* Check that there's just one newline at the end, not two */
    CORRADE_COMPARE(output.find('\n'), output.size() - 1);
```

src/Corrade/Utility/Implementation/WindowsError.cpp
Squareys: ✅

src/Corrade/Utility/CMakeLists.txt
Squareys: ✅

src/Corrade/Utility/Test/DirectoryTest.cpp
Squareys: Yeah, good catch, they are not.

Squareys: ✅

src/Corrade/Utility/Implementation/WindowsError.cpp
Squareys: I learn about `except()` waay too late. So many uses for that

✅

Squareys:
@mosra I fixed the final CI error, awaiting it's completion, then this is ready again.

mosra:
Thanks! :+1:

PR #95, opened 2020-05-04
[MinGW/Clang] Fix SEGFAULTs when logging to globals

Amphaal:
Okay, I think I found the issue with https://github.com/mosra/magnum/pull/417.

Changing few lines in `src/Corrade/Utility/Debug.cpp` allowed me to boot up a sample projet using a Magnum Clang build. 

My 2cts guess here is that Clang does some optimizations and "hard stores" pointer addresses of the standard outputs when directly referenced "as it", references that may change outside of Corrade, resulting in SEGFAULTs when Magnum tries to use its logging capabilities.

### DIFF
--- a/src/Corrade/Utility/Debug.cpp
+++ b/src/Corrade/Utility/Debug.cpp
@@ -117,7 +117,7 @@ CORRADE_VISIBILITY_EXPORT
     #endif
 #endif
 DebugGlobals debugGlobals{
-    &std::cout, &std::cerr, &std::cerr,
+    Debug::defaultOutput(), Warning::defaultOutput(), Error::defaultOutput(),
     #if !defined(CORRADE_TARGET_WINDOWS) ||defined(CORRADE_UTILITY_USE_ANSI_COLORS)
     Debug::Color::Default, false
     #endif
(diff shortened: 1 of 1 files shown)


src/Corrade/Utility/Debug.cpp
```diff
@@ -117,7 +117,7 @@ CORRADE_VISIBILITY_EXPORT
     #endif
 #endif
 DebugGlobals debugGlobals{
-    &std::cout, &std::cerr, &std::cerr,
+    Debug::defaultOutput(), Warning::defaultOutput(), Error::defaultOutput(),
```
mosra, line 120:
```suggestion
    #if defined(CORRADE_TARGET_MINGW) && defined(CORRADE_TARGET_CLANG)
    /* Referencing the globals directly makes MinGW Clang segfault for some reason */
    Debug::defaultOutput(), Warning::defaultOutput(), Error::defaultOutput(),
    #else
    &std::cout, &std::cerr, &std::cerr,
    #endif
```

so it's documented why is it like this (so I don't revert this by accident), and used only when needed. I hope the macros do the right thing -- can you confirm? :)

Amphaal: Nice, works as expected !

Amphaal:
> Wow, nice. I have no idea how you managed to discover this :)

Luck and perseverance I guess... Classic MinGW developper's traits right here xD

mosra:
Merged as abfeeafb3e2a339aa2629148ea400551fbb71987, thank you! (The CIs take ages to go through, sorry.)

PR #98, opened 2020-07-22
make AndroidLogStreamBuffer print to console aswell

mgood7123:
this is useful when executing native executables via a shell

### DIFF
--- a/doc/corrade-changelog.dox
+++ b/doc/corrade-changelog.dox
@@ -161,6 +161,9 @@ namespace Corrade {
     @ref Utility::copy(), @ref Utility::format() and @ref Utility::Sha1 was
     fixed to not call @ref std::memcpy() with @cpp nullptr @ce when size is
     zero (see [mosra/corrade#102](https://github.com/mosra/corrade/issues/102))
+-   @ref Corrade/Utility/StlForwardTuple.h was fixed to do a full
+    @cpp #include <tuple> @ce on libstdc++ before version 7 as there a forward
+    declaration is not available
 
 @subsection corrade-changelog-latest-deprecated Deprecated APIs
 
--- a/src/Corrade/Utility/StlForwardTuple.h
+++ b/src/Corrade/Utility/StlForwardTuple.h
@@ -31,7 +31,7 @@
 @m_since{2019,10}
 
 On @ref CORRADE_TARGET_LIBCXX "libc++", @ref CORRADE_TARGET_LIBSTDCXX "libstdc++"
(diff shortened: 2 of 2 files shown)


src/Corrade/Utility/AndroidLogStreamBuffer.cpp
```diff
@@ -33,7 +34,10 @@ AndroidLogStreamBuffer::~AndroidLogStreamBuffer() = default;
 
 int AndroidLogStreamBuffer::sync() {
     /* Write the data */
-    __android_log_write(std::int32_t(_priority), _tag.data(), str().data());
+    const char * tag = _tag.data();
+    const char * data = str().data();
+    __android_log_write(std::int32_t(_priority), tag, data);
+    std::cout << data;
```
mosra, line 40:
```suggestion
    const std::string data = str();
    __android_log_write(std::int32_t(_priority), _tag.data(), data.data());
    std::cout << data;
```

With the original version, you'd pass a null-terminated array to cout, which means it'd stop on the first null byte (and thus when something goes wrong *real bad*, the message will be cut off). Since there already is a string with size information, just use it.

src/Corrade/Utility/AndroidLogStreamBuffer.cpp
```diff
@@ -33,7 +34,10 @@ AndroidLogStreamBuffer::~AndroidLogStreamBuffer() = default;
...
```
mosra, line 41: This causes *all* output to be put into `std::cout`, with no chance of distincting between stdout or stderr. I'm not sure I like that. Make this configurable? Or choose it based on `_priority`?

Also, what about flushing? Usually when you have error or assertion messages, you'd want them to be printed right away and not in some buffer that might not get emptied before the application crashes / aborts.

mgood7123:
> Hmm, I'm not sure about the use case here -- usually what I have is either a console application, running in a shell (and then I don't need it to fill the android log because I see the output right away), or I have an APK, which doesn't make sense to be run from a shell anyway, and thus filling cout is wasteful as well.
> 
> Unless I missed something, it doesn't look to me like a single app would need both at the same time -- and if it can operate in both contexts, I could imagine it only opting-in to AndroidLogStreamBuffer if it's not run from a shell.
> 
> Furthermore, this doesn't really need a change to this class itself. It could easily be done in a subclass that overrides the `sync()` and then prints to both -- so this means you could easily make a subclass inside your particular poject that has such special needs.

hmmm ok

mgood7123:
how about this?

PR #99, opened 2020-08-15
Directory: Close a leaking handle in list() and test

Squareys:
Hi @mosra !

As per gitter, here's the fix for the leaking handle. Windows will refuse to delete folders with open handles, which means every `list()`-ed folder would be locked until application exit.

Best,
Jonathan

### DIFF
--- a/src/Corrade/Utility/Test/DirectoryTest.cpp
+++ b/src/Corrade/Utility/Test/DirectoryTest.cpp
@@ -40,6 +40,10 @@
 
 #include <clocale>
 
+#ifdef CORRADE_TARGET_WINDOWS
+#include <windows.h>
+#endif
+
 #include "configure.h"
 
 #ifdef CORRADE_UTILITY_LINUX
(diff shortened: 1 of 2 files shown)


src/Corrade/Utility/Test/DirectoryTest.cpp
```diff
@@ -933,9 +937,22 @@ void DirectoryTest::list() {
...
+    #endif
+
     CORRADE_COMPARE_AS(Directory::list(_testDir),
         (std::vector<std::string>{".", "..", "dir", "file"}),
         TestSuite::Compare::SortedContainer);
+
+    #ifdef CORRADE_TARGET_WINDOWS
+    /* Ensure we are not leaking handles */
+    DWORD newHandleCount = 0;
+    GetProcessHandleCount(GetCurrentProcess(), &newHandleCount);
```
mosra, line 952: Wow, this is cool, didn't even know such thing was possible.

You know what? This would be good to do for everything -- I think this could be doable in setup/teardown routines, but I need to patch TestSuite a bit to allow checks there. Is it okay if I merge just the first part with the scope guard so you have the fix and leave this extra check for later when TestSuite is ready?

Squareys: We can also merge this and refactor it once test suite is ready (or refactor with making it ready).

mosra:
Merged the fix alone in 515cc7e31e2c76903a19c52d2b6c82a0f8ccbb96, keeping this open so I have a reminder to add the handle counting to all tests.

Thanks!

PR #114, opened 2021-03-07
Add bazel build definitions

celentes:
#### Easier modularity

Magnum is a modular library that supports a massive amount of different configurations, which complicates setup and build procedure.

Bazel provides better mechanisms for keeping track of distributed dependencies and build configurations than CMake, and I feel like CMake is getting stretched to its limits by this project already.

#### Painless(ish) integration into bigger (non-cpp) projects

E.g. with bazel it is possible to have a devserver running with react page that would automatically watch and recompile magnum-based wasm binary upon any changes in cpp, etc.

#### Better (much better) dependency chain tracking and ease of setup

Bazel fetches dependencies like nothing else does, bringing new devs up to speed with a single command is a big win in my book.

## Implemented
* Corrade linux static/shared builds are implemented and working(ish)
* Set of bazel rules similar to corrade's cmake macros
  - corrade_static_plugin
  - corrade_dynamic_plugin
  - corrade_configure_header (had to write a custom replacement for CMake configure_file mechanism + bazel rule to invoke)

## Not implemented (yet?)
* Static builds with dynamic plugins.
  - I could not, for the life of me, find a way for bazel to export symbols the way cmake does. Plugin and executable both get to have their own sets of global variables.
  - It may not be a big deal because "the bazel way" knows about the entire world at build time and would not have it any other way, meaning that it is the final executable's job to figure out how it wants to be linked in the end. It does not even have a concept of installing a produced library somewhere on the system.
* corrade_add_test
  - It would mostly be a thin macro around bazel's cc_test
  - Bazel does not, currently, have a way of expecting tests to fail, so a shell test wrapper may be in order, but that in turn can well break some things I am too scared to think about.
* Some of the tests
  - See above for tests with expected failure.
  - Some tests like *GlobalStateAcrossLibrariesTest, because I am not at all convinced that those are not false positives with the way bazel links things.
* Crossplatform configs
  - These are planned for.
* Separate static/dynamic configs
  - Bazel has select() mechanism for platforms (os, cpu) and can figure out release/debug compilation modes, but I have absolutely no idea how to make an extra define on a static build. Welp.

## Further goals
* Crossplatform selects for bazel
* Easy emscripten build with zero setup
* Bazel build definitions for examples
* Bazel build definitions for magnum 

## Nongoals
* Replacing CMake :D

### Previously attempted

I have implemented a working configuration with bazel calling CMake to assemble magnum before, and that works reasonably well even in a zero-previous-setup scenario (i.e. with just docker, jump into a container, press one button - bazel fetches ninja, cmake, compiles both, fetches emscripten, compiles all of magnum/corrade/dependencies to wasm).

That still has some pain points:
* It does not reuse bazel's toolchain resolution mechanisms well
* Does not play nice with remote cache
  - Calling cmake is a single huge invocation that produces similarly large cache artifact.
* Does not play well with bazel's work scheduler threading
  - Bazel thinks one invocation lives in one thread, so either you have slow builds or they start to fight with ninja.
* About 40s of overhead on each CMake invocation
  - I cannot explain this.

### Pain points
* Corrade's tight coupling with CMake
* Opinionated bazel
  - Bazel's linking mechanism works really well with buy-in, but is otherwise fairly obscure, hence the static build + dynamic plugins conundrum
  - It is fairly difficult doing anything out of the ordinary because of bazel's (pretty anal) restrictions on what you can and cannot do with the repo, and how you can actually query various bits of info during the build time.
  - Seriously, no static/dynamic defines?
* Bazel documentation
  - There is plenty of thereof, but it is not straightforward, not complete and not always very helpful. Not dissimilar to trying to find a way out of a maze while blind, deaf and drunk.
* Some Corrade tests and mechanisms have assumptions that are broken by bazel - mostly filesystem access and how bazel sandboxes things and how corrade does not know about it.
  - Currently, UtilityDirectoryTest and PluginManagerManagerTest are failing some checks.


mosra:
Huh, what a huge diff :) I need to think about this a bit more and will write a detailed reply later, but here's my initial thoughts:

- I have no resources to maintain another buildsystem in addition to CMake, sorry. This has to live in a separate project. I'm trying to not be too tied to CMake (except for generating the `configure.h` file, which has to be done _somehow_) to make it possible for people to use whatever they want, but for the last three years there's constantly *more* work than my time, and I'm working on these projects full time. That means I have to prioritize and if I should choose between maintaining one extra buildsystem or spending that time writing a better data processing pipeline, guess what I pick :)
- I can however work on minimizing some of the pain points so it's easier to not use CMake (for example the `corrade_add_resource()` macro generates dynamic depenendencies parsed from the resource configuration file and currently that's done with a *nasty* CMake macro, which could be provided by `corrade-rc` itself instead and be usable also outside of CMake).
- Unless *you specifically* want to run tests in Bazel, I don't think these need to be anywhere else than in CMake. That should also reduce the complexity of this alternative buildsystem quite a lot, because you no longer need to build certain files twice for testing assertions, pass various environment / compile flags parameters to each, bundle/whitelist certain filesystem locations so tests can see them etc.

Expanding on the first point a bit -- I was convinced into maintaining a Bazel buildsystem "clone" for all Magnum projects by one company few years ago and, to be polite, it was *not great*, especially when it came to external dependencies. Corrade and Magnum are ~fine, but when you get to the plugins, where every second plugin has a 3rd party dependency, it's *hell*. I don't know how is it nowadays, but back then the workflow was basically to recursively port everything to Bazel (libjpeg, libpng, zlib, libxml2, assimp, faad2 ... you name it), and it was just a neverending maintenance nightmare with half the things breaking on every second bazel upgrade and a lot of corners cut ("let's pipe our output to a ffmpeg binary instead of trying to build it ourselves, it will save us *weeks*") and for every hour spent actually coding, it was *two days* of trying to reach out to some bazel guru to fix a blocker build issue. Never again, I don't want to deal with that thing *ever again* myself. If I would be Google and would have *all my dependencies* in Bazel already, then maybe this would make sense, but not when I need to depend on 50+ (and counting) other libraries with each having their own buildsystem, each having a different release schedule with each release having the build broken in a different way. I have enough pain with just getting *their own* buildsystem to work well enough on platforms I need, I can't imagine recursively rewriting these with Bazel.

celentes:
Was thinking it might be a point, pushing more maintenance your way.
Nowadays bazel has a repo with rules to invoke external build systems, so as long as it's in cmake/make/somethingsensible, it's not too dreadful, but i'm not gonna try to convince you into it :)

I can rewrite these to keep it in the outside repo.

Would be good to get the couple of failing tests into green, it was mostly stuff like DirectoryTest trying to look up corrade-rc where it expects it. Like, if Directory::isSandboxed() could detect it's being run by bazel. I mostly convert tests to see if the buildsystem switch maintains sanity checks.

celentes:
After about an hour of vimmery, I moved the entire thing to [external repo](https://github.com/celentes/corrade-bazel).
None of the original sources are in it, but i'd like to keep it under the original license. Should I?

I'll play about with whatever failing tests left and see if I can submit a reasonable pull request to cover those.

mosra:
Closing this old PR for the reasons stated above, in particular the maintenance overhead of maintaining two separate buildsystems.

Issue #116, opened 2021-04-29
ArrayTuple: Initialization of larger structures


xqms:
Started from https://github.com/mosra/magnum/issues/505. Here's a reproducing example using just Corrade:

```c++
#include <Corrade/Containers/ArrayTuple.h>
#include <Corrade/Containers/ArrayView.h>
#include <Corrade/Utility/Debug.h>

using namespace Corrade;

int main(int argc, char** argv)
{
    Containers::ArrayView<std::pair<unsigned int, unsigned int>> arr;

    Containers::ArrayTuple tuple{
        {Containers::ValueInit, 4, arr}
    };

    for(std::size_t i = 0; i < arr.size(); ++i)
    {
        Utility::Debug{} << i;
        if(arr[i].second != 0)
            Utility::Debug{} << "Yes!";
    }

    return 0;
}
```

Valgrind gives warnings about uninitialized memory after half of the array. I believe the problem is this line:

https://github.com/mosra/corrade/blob/1114af582e714fc47979b5e847a68572ec0589b7/src/Corrade/Containers/ArrayTuple.cpp#L280-L283

where it should be `items[i]._constructor(_data + offset + j*items[i]._elementSize);` - at least that fixes the valgrind errors.

Probably a unit test with larger structures is required, that tests whether the memory has been correctly initialized.

mosra:
Wow, that was a stupid mistake from me. Thanks for all the investigation :+1:

PR #117, opened 2021-04-30
Fix ArrayTuple element construction

xqms:
Fix for #116. It also includes a unit test which exposes the problem.

Curiously, I couldn't get ASan to detect the usage of uninitialized memory. Valgrind reports it, however.

### DIFF
--- a/src/Corrade/Containers/Test/ArrayTupleTest.cpp
+++ b/src/Corrade/Containers/Test/ArrayTupleTest.cpp
@@ -55,6 +55,8 @@ struct ArrayTupleTest: TestSuite::Tester {
     void constructCopy();
     void constructMove();
 
+    void constructBig();
+
     void allocatorAlignmentEmpty();
     template<int a> void allocatorAlignmentFromItems();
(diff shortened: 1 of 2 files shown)


src/Corrade/Containers/Test/ArrayTupleTest.cpp
```diff
@@ -651,6 +655,80 @@ void ArrayTupleTest::constructMove() {
...
+{
+    Big::constructed = 0;
+    Big::destructed = 0;
+
+    {
+        ArrayView<char> chars;
+        ArrayView<Big> bigs;
+        ArrayTuple data{
+            {17, chars},
+            {7, bigs}
```
mosra, line 697: `bigs`, haha :)

mosra (approved):
The unit test is great, thanks.

src/Corrade/Containers/Test/ArrayTupleTest.cpp
xqms: well, we have to be consistent :D

PR #120, opened 2021-06-18
Fix typo in Containers::arraySize doc snippet

pezcode:
I have to admit I still have the tiniest bit of fear that this is some arithmetic going over my head


mosra:
> this is some arithmetic going over my head

No, it just looks like I didn't press <kbd>1</kbd> hard enough :D Thanks.

PR #125, opened 2021-10-17
RPM build

1b00:
To build RPM packages into the default location cd into package/rpm, read the docs and run build_rpm.sh (with or without parameter: pass either 'master' or any valid git tag|hash or no parameter for last tag)


mosra:
Sorry for neglecting your contribution for so long. This was now combined with #186 and merged as 9ef83b7291371900fd24c307bac66433941f2114.

PR #127, opened 2022-02-19
std::tuple_element / tuple_size specializations for Corrade containers

sthalik:
This commit adds `std::tuple_element` and `std::tuple_size` forward declarations to avoid slurping in thousands of lines worth of STL headers. For unsupported STL versions, `<array>` is included instead.

### DIFF
--- a/src/Corrade/Containers/OptionalStl.h
+++ b/src/Corrade/Containers/OptionalStl.h
@@ -36,6 +36,7 @@ information.
 */
 
 #include <optional>
+#include <type_traits>
 
 #include "Corrade/Containers/Optional.h"
 
@@ -64,6 +65,31 @@ template<class T> struct OptionalConverter<T, std::optional<T>> {
 template<class T> struct DeducedOptionalConverter<std::optional<T>>: OptionalConverter<T, std::optional<T>> {};
 
 }}}
+#endif
+
+namespace std {
+
+template<class T> struct tuple_size<Corrade::Containers::Optional<T>> : std::integral_constant<std::size_t, 2> {};
+template<class T> struct tuple_element<0, Corrade::Containers::Optional<T>> { using type = T; };
+template<class T> struct tuple_element<1, Corrade::Containers::Optional<T>> { using type = bool; };
+
+}
+
+#if CORRADE_CXX_STANDARD >= 201703
+namespace Corrade { namespace Containers {
+
+template<std::size_t N, class T>
+typename std::tuple_element<N, Optional<T>>::type
+get(const Optional<T>& value) noexcept(std::is_nothrow_default_constructible<T>() && std::is_nothrow_copy_constructible<T>())
+{
+    static_assert(N < 2, "");
+    static_assert(std::is_default_constructible<T>() && std::is_copy_constructible<T>(), "");
+    if constexpr (N == 0)
+        return value ? *value : T{};
+    if constexpr (N == 1)
+        return bool(value);
+}
+}}
 #endif
 
 #endif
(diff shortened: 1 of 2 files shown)


src/Corrade/Containers/Test/CMakeLists.txt
```diff
@@ -69,6 +69,7 @@ corrade_add_test(ContainersStringViewTest StringViewTest.cpp LIBRARIES CorradeUt
 corrade_add_test(ContainersStringViewStlTest StringViewStlTest.cpp)
 corrade_add_test(ContainersTripleTest TripleTest.cpp)
 corrade_add_test(ContainersTripleStlTest TripleStlTest.cpp)
+corrade_add_test(ContainersTupleStlTest TupleStlTest.cpp)
```
mosra, line 72: To make the tests on ancient compilers stop failung, move this inside the C++17 `if()` below, same place where you set the C++17 standard

mosra:
Since I'm deep in the reviews already, let's unblock this too, *finally* (it's from February? wow :sweat_smile:).

As the change now boils down to `#include <utility>` to get declarations of `std::tuple_element` and `std::tuple_size`, I don't think it's needed to invent any special forward-declarating header for these two -- the size of `<utility>` is rather small, while maintenance cost of the forward-declaring header and potential for strange breakages after compiler upgrades would be far higher than that. So just drop the `TupleStl.h` altogether.

What's left is that I assume you'd want to use C++17 `auto [a, b] = someType;` for as many Corrade/Magnum types as possible, right? In case of Corrade that'd be mainly these I suppose:

- `Containers::Pair`, for which the `std::tuple_whatever` specializations could go into the already-existing `Containers/PairStl.h`, and a corresponding new C++17-only `PairStlCpp17Test.cpp`
- `Containers::Triple`, for which the `std::tuple_whatever` specializations could go into the already-existing `Containers/TripleStl.h`, and a corresponding new C++17-only `TripleStlCpp17Test.cpp`
- `Containers::StaticArrayView`, for which it could go into the already-existing `Containers/ArrayViewStl.h`, and a corresponding new `ArrayViewStlCpp17Test.cpp`

For completeness there's also `Containers::StaticArray` but it doesn't have a corresponding `StaticArrayStl.h` header yet, that one you'd have to create, and a new `StaticArrayStlCpp17Test.cpp` as well. (It's also fine if you implement just the `Pair` and `Triple`, those are the most important ones.)

It seems like a lot of new test files, but my assumption is that there will be more and more C++14/17/20-specific features (such as your constexpr additions), and the new test files will be reused for those as well. 

The `std::tuple_element` / `std::tuple_size` is a C++11 feature already so I wouldn't hide them in any `#if compiling_as_cpp17`. Having them exposed unconditionally would mean `std::get<N>()` working with Corrade/Magnum types even under C++11, for example.

In case of Magnum I assume you'd want `Math::Vector` exposed this way, so that'd be a new `VectorStl.h`, and a corresponding new C++17-only `VectorStlCpp17Test.cpp`.

sthalik:
> The std::tuple_element / std::tuple_size is a C++11 feature already so I wouldn't hide them in any #if compiling_as_cpp17. Having them exposed unconditionally would mean std::get<N>() working with Corrade/Magnum types even under C++11, for example.

For types without an already-existing `get<N>` member function, this makes the implementation uglier since `if constexpr` is from C++17 and functions can't be partially specialized.

mosra:
So, not going to do the specializations for other containers either? I'll convert that to an issue, then, in case it eventually becomes useful for someone else.

sthalik:
I still plan on doing it for Pair and Triple at least, but have more pressing issues at this time.

mosra:
For anyone still subscribed to this PR, a variant of the above together with structured bindings for `StridedDimensions` was added in 5f4fc52ad2f07c4e022e8ba8e770144becb69b17.

PR #137, opened 2022-05-26
Small micro-optimizations for Sha1

Squareys:
Hi @mosra,

as mentioned on gitter, here's the minor optimizations I did. The only really significant one is 3b0bce3352455e095bca9b2130dd0c5fada4472d, in case you want to cherry-pick it for minimal invasiveness.

Best,
Jonathan

### DIFF
--- a/src/Corrade/Utility/Sha1.cpp
+++ b/src/Corrade/Utility/Sha1.cpp
@@ -120,8 +120,8 @@ Sha1::Digest Sha1::digest() {
     _bufferSize += 8;
 
     /* Process remaining chunks */
-    for(std::size_t i = 0; i != _bufferSize/64; ++i) {
-        processChunk(_buffer+i*64);
+    for(std::size_t i = 0; i != _bufferSize; i += 64) {
+        processChunk(_buffer + i);
     }
 
     /* Convert digest from big endian */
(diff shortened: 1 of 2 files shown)


Squareys:
No worries, we have since switched to [xxhash3](https://github.com/pombredanne/xxHash-3), which has insanely high throughput.

mosra:
I needed to avoid an `<algorithm>` include, remembered this PR existed and cherry-picked most of it as 6492b311979d8eefaa437f2220d4307779bb6e5c...d1f6f4affda3355bd67a796e2babdeb5eb992fb4.

When I fixed the benchmark to test the actual hot loop and not the Sha1 instance construction or `std::string` allocation from data that were different from the intended `randomData` ( :sweat_smile: ) the Release measurements weren't really different before and after the loop split, so I was considering not even including that change because it wasn't worth the code size expansion. The split helped in Debug only if `CORRADE_ALWAYS_INLINE` was added as well (~20% faster), otherwise not at all. Which ... I guess is a valid improvement, so ultimately I added that part as well. With `CORRADE_ALWAYS_INLINE` alone the Debug build improvement was about 10%.

I might get to adding the x86 instructions eventually but as you say, there are better hashes nowadays, so probably not worth the time investment anymore.

PR #142, opened 2022-06-11
Utility: more argv const in Arguments::parse

sthalik:


src/Corrade/Utility/Arguments.h
```diff
@@ -873,7 +863,7 @@ class CORRADE_UTILITY_EXPORT Arguments {
          * additional arguments are parsed, only `--help` option is set and
          * @cpp true @ce is returned.
          */
-        bool tryParse(int argc, const char** argv);
+        bool tryParse(int argc, const char* const* argv);
 
         /** @overload */
         bool tryParse(int argc, char** argv) {
```
mosra, line 869: Can you drop these as well? Thank you! :blush:

mosra:
PR with a net negative line count is never bad :D Thank you. 

Now I need to look where else I forgot to realize `const char* const*` is a thing,

PR #144, opened 2022-06-11
Utility: update Path::{home,configuration}Directory() on Windows.

williamjcm:
Instead of the old `SHGetFolderPathW()` API, they now use the `SHGetKnownFolderPath()` API, which was added in Windows Vista.

As a side effect, if OneDrive is installed, `homeDirectory()` should now return the proper Documents directory, instead of the OneDrive equivalent. Though I couldn't test that, because I made sure to remove it from my machine as soon as possible.

### DIFF
--- a/src/Corrade/Utility/Path.cpp
+++ b/src/Corrade/Utility/Path.cpp
@@ -594,11 +594,18 @@ Containers::Optional<Containers::String> homeDirectory() {
 
     /* Windows (not Store/Phone) */
     #elif defined(CORRADE_TARGET_WINDOWS) && !defined(CORRADE_TARGET_WINDOWS_RT)
-    /** @todo get rid of MAX_PATH */
-    wchar_t h[MAX_PATH + 1];
+    wchar_t* h = nullptr;
+    Containers::ScopeGuard guard{h,
+        #ifdef CORRADE_MSVC2015_COMPATIBILITY
+        /* MSVC 2015 is unable to cast the parameter for CoTaskMemFree */
+        [](LPVOID path){ CoTaskMemFree(path); }
+        #else
+        CoTaskMemFree
+        #endif
+    };
     /* There doesn't seem to be any possibility how this could fail, so just
        assert */
-    CORRADE_INTERNAL_ASSERT(SHGetFolderPathW(nullptr, CSIDL_PERSONAL, nullptr, 0, h) == S_OK);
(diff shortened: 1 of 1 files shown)


mosra:
Ignore the Apple CI failures, I cancelled them because they just wait forever (5d6fed8bcd0136eed594ffb16308e0c3a0b5c441 will fix that). 

Let's see what AppVeyor says, then I'll merge.

src/Corrade/Utility/Path.cpp
```diff
@@ -640,11 +647,18 @@ Containers::Optional<Containers::String> configurationDirectory(const Containers
 
     /* Windows (not Store/Phone) */
     #elif defined(CORRADE_TARGET_WINDOWS) && !defined(CORRADE_TARGET_WINDOWS_RT)
-    /** @todo get rid of MAX_PATH */
-    wchar_t path[MAX_PATH];
+    wchar_t* path = nullptr;
+    Containers::ScopeGuard guard{path,
+        #ifdef CORRADE_MSVC2015_COMPATIBILITY
+        /* MSVC 2015 is unable to cast the parameter for CoTaskMemFree */
+        [](LPVOID path){ CoTaskMemFree(path); }
```
mosra, line 654: I think 

```suggestion
        [](wchar_t* path){ CoTaskMemFree(path); }
```

could fix the MSVC2015 failures? Same in the other case.

williamjcm: Oh, sorry I didn't notice it earlier. I'll fix that. 😅

mosra:
To fix MinGW, this could help I think.

```patch
diff --git a/src/Corrade/Utility/Path.cpp b/src/Corrade/Utility/Path.cpp
index 5a365189..08646468 100644
--- a/src/Corrade/Utility/Path.cpp
+++ b/src/Corrade/Utility/Path.cpp
@@ -74,6 +74,10 @@
 /** @todo remove the superfluous includes when mingw is fixed (otherwise causes undefined EXTERN_C error) */
 #ifdef CORRADE_TARGET_WINDOWS
 #ifdef __MINGW32__
+/* For SHGetKnownFolderPath() on MinGW: https://stackoverflow.com/a/59033437 */
+#ifndef NTDDI_VERSION
+#define NTDDI_VERSION 0x06000000
+#endif
 #include <wtypes.h>
 #include <windef.h>
 #include <wincrypt.h>
```

And if you rebase on master, the Mac CI stops being cursed.

mosra:
Merged as 20c1ba383753c205d1f48c66f50b112e6d9490fd, with the MinGW patch applied as 612c49a1eb1296a9ce7e267def93d1b272ae95d2. Thank you for the fix! :+1:

PR #146, opened 2022-08-10
Utility: use Corrade String and StringView in ConfigurationValue

hugoam:
Hello !
We're trying to optimize our compile times and I noticed that `StlForwardString.h` is actually very heavy on MSVC because it just directly includes `<string>`, so I thought let's go and try to clean up  `ConfigurationValue`.
I'll open the associated Magnum PR right away. Let me know if I'm going at this correctly 😅

### DIFF
--- a/doc/snippets/Utility.cpp
+++ b/doc/snippets/Utility.cpp
@@ -75,11 +75,11 @@ namespace Corrade { namespace Utility {
 template<> struct ConfigurationValue<Foo> {
     static std::string toString(const Foo& value, ConfigurationValueFlags flags) {
         return
-            ConfigurationValue<int>::toString(value.a, flags) + ' ' +
+            ConfigurationValue<int>::toString(value.a, flags) + " " +
             ConfigurationValue<int>::toString(value.b, flags);
     }
 
-    static Foo fromString(const std::string& stringValue, ConfigurationValueFlags flags) {
+    static Foo fromString(Containers::StringView stringValue, ConfigurationValueFlags flags) {
         std::istringstream i{stringValue};
         std::string a, b;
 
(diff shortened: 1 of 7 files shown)


mosra:
Hi and sorry for a late reply -- I've been away last week and only now catching up to everything.

This makes sense in general, and especially the new `ConfigurationValueStl.h` header does. That's how I'd do it myself as well. But the reason I neglected this part was because all APIs that *actually used* `ConfigurationValue` use `std::string` themselves heavily, so until I clean those up I didn't really see how fixing it just here would improve anything. Can you say where it helped in your case?

Oh, or is it just due to `Mesh.h`, `PixelFormat.h` and `VertexFormat.h` dragging `<string>` in (on MSVC), according to what I see in mosra/magnum#582? Because there it relies just on a forward declaration of `ConfigurationValue` and could thus be changed to `String` / `StringView` without needing to modify anything else. The `Math/ConfigurationValue.h` is a separate header that isn't pulled in by anything else, so it shouldn't affect your build times.

(For the rest, I feel it wouldn't really improve anything until `Utility::Arguments` and `Utility::Configuration` are de-STL-string-ified first. So I'd keep these PRs around until I get to cleaning that up.)

hugoam:
>Oh, or is it just due to Mesh.h, PixelFormat.h and VertexFormat.h dragging <string> in (on MSVC)

Yes, exactly, that was the low hanging fruit I was trying to tackle to improve compile times as those are some pretty commonly included headers, and have them drag `<string>` induce a pretty big cost. So yep downsizing this to just changing those would be perfectly fine for us ! Unless obviously there is something else that we commonly include that I missed

mosra:
Done! Leaving the PRs open, will merge the rest once I get to cleaning up `Utility::Arguments` and `Utility::Configuration`.

PR #147, opened 2022-08-14
Add GDB pretty-printers for some types.

williamjcm:
As mentioned on Gitter, here's the PR adding adding GDB pretty-printers

Most types in `Containers`, as well as most `Utility::Configuration*` types (except values, since the default printing for them is adequate), are supported.

For multi-dimensional `StridedArrayView`, since I couldn't manage to get nested printing working, flattened printing is used instead.

The only `Containers` type to not be supported is `ScopeGuard`, because type erasure makes it hard to pretty-print its contents.

Note: because the `gdb.Type.template_argument()` method chokes on non-type arguments (such as the size used by `StaticArray(View)`, or `StridedArrayView`'s dimensions) despite [this bug supposedly being fixed in GDB 7.2](https://sourceware.org/bugzilla/show_bug.cgi?id=11060), I had to rely on string manipulation to extract those arguments.

### DIFF
--- a/src/debuggers/CMakeLists.txt
+++ b/src/debuggers/CMakeLists.txt
@@ -4,6 +4,7 @@
 #   Copyright © 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016,
 #               2017, 2018, 2019, 2020, 2021, 2022
 #             Vladimír Vondruš <mosra@centrum.cz>
+#   Copyright © 2022 Guillaume Jacquemin <williamjcm@users.noreply.github.com>
 #
 #   Permission is hereby granted, free of charge, to any person obtaining a
 #   copy of this software and associated documentation files (the "Software"),
@@ -24,25 +25,5 @@
 #   DEALINGS IN THE SOFTWARE.
 #
 
-# On MSVC remove /W3, as we are replacing it with /W4. Could be removed as of
-# 3.15 with this: https://cmake.org/cmake/help/latest/policy/CMP0092.html
-if(CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" OR CMAKE_CXX_SIMULATE_ID STREQUAL "MSVC")
-    string(REPLACE "/W3" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
-endif()
-
-set_directory_properties(PROPERTIES
-    CORRADE_CXX_STANDARD 11
-    CORRADE_USE_PEDANTIC_FLAGS ON)
-
-# On Windows enable UNICODE/_UNICODE macros to avoid accidentally passing UTF-8
-# values to ANSI functions
-if(CORRADE_TARGET_WINDOWS)
-    add_definitions("-DUNICODE" "-D_UNICODE")
-endif()
-
-add_subdirectory(Corrade)
-
-# Corrade include dir for superprojects
-set(CORRADE_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR} CACHE INTERNAL "" FORCE)
-
-add_subdirectory(debuggers)
+install(DIRECTORY gdb
+    DESTINATION ${CORRADE_DATA_INSTALL_DIR}/debuggers)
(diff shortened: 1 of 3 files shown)


mosra:
Thank you! I looked over the code and everything seems fine, even the coding style is exactly as I'd do it myself, haha. I don't really use gdb myself so I can't *prove* it actually works, but I think I can trust you after all these years :)

Two questions before I merge:

- What should I put into the docs so people know how to use this?
- I was wondering if it would make sense to restrict the installation only if GCC/Clang is used (i.e., don't install if MSVC is used). But then there's a case of MinGW, and clang-cl, and maybe GDB can work with PDB files, who knows, and I'm not sure if all that extra logic is worth the bother, so probably just easiest to install always?

> flattened printing is used instead.

That's fine. It can be always added later if someone figures out how.

> ScopeGuard

This is fine as well, I don't see this type being too important. Besides that, the type usually just takes a value/variable that's stored somewhere else already, so the user can print the other one instead.

> chokes on non-type arguments

Ah, sigh. So I suppose the Magnum math types all need the same workaround as well.

williamjcm:
> - What should I put into the docs so people know how to use this?

Oh, boy... :D

To be honest, that's actually the complex part because, once again, there's no easy way of loading pretty-printers.

It's possible to do this:
1. manually through GDB's commands
2. by sourcing a GDB script file (there's a handy, well, `source` command for that)
3. by enabling auto-loading of scripts ([GDB docs](https://sourceware.org/gdb/onlinedocs/gdb.html#Initialization-Files), [JetBrains CLion docs](https://www.jetbrains.com/help/clion/configuring-debugger-options.html#gdbinit-lldbinit)).
  - In the case of MSYS2/MinGW's GDB, the per-user `.gdbinit` file can be either in the MSYS2 `$HOME/.gdbinit`, or in `C:\Users\<name>\.config\gdb\gdbinit`, depending on whether it's run in a MSYS2 shell or not. Non-MSYS2 GDB might use the latter too, but I currently don't have one of these to test with.

The scripts themselves tend to be a bit convoluted and a bit "WTF". Here's what GDB on MinGW-w64/MSYS2 does when loading the libstc++ pretty-printers (`/mingw64/etc/gdbinit`), for example:

```text
python
import sys
sys.path.insert(0, sys.path[0] + '/../../gcc-12.1.0/python')
from libstdcxx.v6.printers import register_libstdcxx_printers
register_libstdcxx_printers (None)
end
```

Loading the Corrade pretty-printers might look like this (skipping the `python` and `end` commands for the sake of syntax highlighting:

```py
import sys
sys.path.insert(0, '/path/to/share/') # not pointing to the debuggers subfolder here because it could conflict with the built-in gdb module
from corrade.debuggers.gdb import register_corrade_printers
register_corrade_printers(gdb.current_objfile())
```

Initialisation scripts for an "objfile" (aka a binary or dynamic library) can be directly a Python script, as per [GDB's docs](https://sourceware.org/gdb/onlinedocs/gdb/Python-Auto_002dloading.html#Python-Auto_002dloading), in which case the `python` and `end` commands aren't needed.

It's also possible to register printers to a "progspace" (aka the entire inferior, including loaded libraries) using `gdb.current_progspace()`, though I don't know if it *actually* makes a difference. Passing `None` works too, and will register the pretty-printers globally, which may be unsuitable when dealing with multiple inferiors, as the printers could get registered multiple times.

That's all I could think of at the moment. If I missed anything that could be of use, don't hesitate to mention it.

> - I was wondering if it would make sense to restrict the installation only if GCC/Clang is used (i.e., don't install if MSVC is used). But then there's a case of MinGW, and clang-cl, and maybe GDB can work with PDB files, who knows, and I'm not sure if all that extra logic is worth the bother, so probably just easiest to install always?

MinGW being pretty much GCC, I don't think this should be an issue.

In the case of Clang, it depends on whether GDB or LLDB is used to debug, and LLDB uses its own API according to [the official docs](https://lldb.llvm.org/use/variable.html#python-scripting), meaning the GDB scripts are useless in this case.

IIRC, clang-cl generates PDB debug info, meaning only VS (and LLDB, thanks to [MS being more open about the format](https://github.com/Microsoft/microsoft-pdb)) can be used.

GDB can't work with PDB files yet, but it's possible the option might come in the future, like with LLDB above.

mosra:
Awesome, that's quite an exhaustive info. Will merge tomorrow or later this week, thank you!

mosra:
Merged as 43d8fa0dae83e0542c919c5b90614c8215ee7865, finally. Sorry for the extreme delay and thanks!

PR #148, opened 2022-09-07
Add missing noexcept for some move ctors/assigns

Skylion007:
I applied some fixes with the latest clang-tidy tools to Corrade. Mainly some missing noexcept specifies for move assignments, which will allow it to use move operations in STL containers. 

LinkedLIst could also have some methods marked as noexcept, although they would not actually be noexcept with CORRADE_ASSERTS (in debug mode). They are noexcept in Release Mode though.

### DIFF
--- a/src/Corrade/Utility/Configuration.cpp
+++ b/src/Corrade/Utility/Configuration.cpp
@@ -72,14 +72,14 @@ Configuration::Configuration(std::istream& in, const Flags flags): Configuration
     if(parse({data.data(), data.size()})) _flags |= InternalFlag::IsValid;
 }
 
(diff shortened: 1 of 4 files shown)


Skylion007:
@mosra This PR should be good to go.

mosra:
Yeah I was neglecting these because the whole Configuration API is due for a serious redesign :) Thanks in any case.

> they would not actually be noexcept with CORRADE_ASSERTS (in debug mode)

Just to correct here a bit -- the asserts are not exceptions (they're aborting), and thus can appear in `noexcept` functions. And `CORRADE_ASSERT()` stays there even in Release, the `CORRADE_DEBUG_ASSERT()` macros are debug-build-only.

mosra:
Argh GitHub UI what are you doing. Merged as 1d214f6a182880b4316a5bf315d6eaf3293f8e64 :)

PR #151, opened 2022-10-04
fix MSVC warning with -W4

sthalik:
In `Containers::BitArray`, cast `bool` to `char` before performing arithmetic on it.


pezcode:
The same warning exists in `BitArrayView`, should probably be fixed there as well

mosra:
Thank you both! :+1:

PR #152, opened 2022-10-09
Add macros for conditional constexpr

sthalik:
This is going to be used in Magnum vectors.

### DIFF
--- a/src/Corrade/Utility/Test/CMakeLists.txt
+++ b/src/Corrade/Utility/Test/CMakeLists.txt
@@ -234,6 +234,15 @@ if(NOT CMAKE_CXX_FLAGS MATCHES "-std=")
         corrade_add_test(UtilityMacrosCpp20Test MacrosCpp20Test.cpp)
         set_target_properties(UtilityMacrosCpp20Test PROPERTIES CORRADE_CXX_STANDARD 20)
     endif()
+
+
+    if((CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS "9.0") OR
+       (CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS "9.0") OR
+       (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS "11.4") OR
+       (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" AND NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS "19.25"))
+        corrade_add_test(UtilityTypeTraitsCpp20Test TypeTraitsCpp20Test.cpp)
+        set_target_properties(UtilityTypeTraitsCpp20Test PROPERTIES CORRADE_CXX_STANDARD 14)
+    endif()
 endif()
 
 # It should return with a non-zero exit code
--- a/src/Corrade/Utility/Test/TypeTraitsCpp20Test.cpp
+++ b/src/Corrade/Utility/Test/TypeTraitsCpp20Test.cpp
@@ -1,0 +1,68 @@
+/*
+    This file is part of Corrade.
+
+    Copyright © 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016,
+                2017, 2018, 2019, 2020, 2021, 2022
+              Vladimír Vondruš <mosra@centrum.cz>
+
+    Permission is hereby granted, free of charge, to any person obtaining a
+    copy of this software and associated documentation files (the "Software"),
+    to deal in the Software without restriction, including without limitation
+    the rights to use, copy, modify, merge, publish, distribute, sublicense,
+    and/or sell copies of the Software, and to permit persons to whom the
+    Software is furnished to do so, subject to the following conditions:
+
+    The above copyright notice and this permission notice shall be included
+    in all copies or substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+    IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+    THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+    LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+    FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+    DEALINGS IN THE SOFTWARE.
+*/
+
+#include "Corrade/TestSuite/Tester.h"
+#include "Corrade/Utility/TypeTraits.h"
+#include <cstdio>
+
+namespace Corrade { namespace Utility { namespace Test { namespace {
+
+struct TypeTraitsCpp20Test: TestSuite::Tester {
+    explicit TypeTraitsCpp20Test();
+    void isConstantEvaluatedTest();
+};
+
+TypeTraitsCpp20Test::TypeTraitsCpp20Test() {
+    addTests({&TypeTraitsCpp20Test::isConstantEvaluatedTest});
+}
+
+#ifdef CORRADE_IS_CONSTANT_EVALUATED
+constexpr int constevalHelper(int i) {
+    if (CORRADE_IS_CONSTANT_EVALUATED)
+        return i + 1;
+    else {
+        std::printf("");
+        return i + 2;
+    }
+}
+#endif
+
+void TypeTraitsCpp20Test::isConstantEvaluatedTest() {
+#ifndef CORRADE_IS_CONSTANT_EVALUATED
+    CORRADE_SKIP("CORRADE_IS_CONSTANT_EVALUATED not supported on this compiler.");
+#else
+    constexpr int retConstant = constevalHelper(0);
+    static_assert(retConstant == 1, "");
+    CORRADE_COMPARE(retConstant, 1);
+    const volatile int arg = 0;
+    const int retRuntime = constevalHelper(arg);
+    CORRADE_COMPARE(retRuntime, 2);
+#endif
+}
+
+}}}}
+
+CORRADE_TEST_MAIN(Corrade::Utility::Test::TypeTraitsCpp20Test)
--- a/src/Corrade/Utility/TypeTraits.h
+++ b/src/Corrade/Utility/TypeTraits.h
@@ -168,6 +168,26 @@ template<class U> class className {                                         \
         enum: bool { value = sizeof(get(std::declval<U>())) == sizeof(char) }; \
 }
 
+/** @hideinitializer
+@brief C++20 is_constant_evaluated
+@m_since{2022,10}
+
+Expands to a predicate determining whether given @cpp constexpr @ce function
+is being evaluated at compile-time or not. Under C++14 rules, constexpr
+functions may be defined as long as at least some of the code paths are able
+to be executed at compile-time. As long as a fallback is present, features
+such as SIMD or inline assembly may be used in functions marked constexpr.
+
+This support is available on all C++20 capable compilers, but also certain
+ones (Clang 9, GCC 9, MSVC 2022 17.1) that expose the feature as a non-portable
+extension. In which case it may be used under C++14 relaxed constexpr rules.
+*/
+#if (defined(CORRADE_TARGET_CLANG) && !defined(CORRADE_TARGET_APPLE_CLANG) && __clang_major__ >= 9) || (defined(CORRADE_TARGET_APPLE_CLANG) && __clang_major__*100 + __clang_minor__ >= 1104) || (defined(CORRADE_TARGET_GCC) && __GNUC__ >= 9) || (defined(CORRADE_TARGET_MSVC) && _MSC_VER >= 1931)
+#define CORRADE_IS_CONSTANT_EVALUATED (__builtin_is_constant_evaluated())
+#elif CORRADE_CXX_STANDARD >= 202002
+#define CORRADE_IS_CONSTANT_EVALUATED (std::is_constant_evaluated())
+#endif
+
 namespace Implementation {
     /* As of Eigen 3.4.0, due to these two commits in particular,
         https://gitlab.com/libeigen/eigen/-/commit/c0ca8a9fa3e03ad7ecb270adfe760a1bff7c0829


src/Corrade/configure.h.cmake
```diff
@@ -403,4 +403,22 @@ static_assert(sizeof(1 ? "" : "") == 1,
...
+#define CORRADE_CXX17_CONSTEXPR constexpr
+#else
+#define CORRADE_CXX17_CONSTEXPR
+#endif
+
+#if CORRADE_CXX_STANDARD >= 201402L
+#define CORRADE_CXX14_CONSTEXPR constexpr
+#else
+#define CORRADE_CXX14_CONSTEXPR
+#endif
```
mosra, line 422: There's already [CORRADE_CONSTEXPR14](https://doc.magnum.graphics/corrade/Macros_8h.html#af57746d07f1b55515249dd8f7f1d9e11) in `Corrade/Utility/Macros.h`, and I'd like the other two to follow a similar naming and be in the same header.

But also please see my comment in mosra/magnum#597 about having a special macro for C++14+ constexpr only if a "is constant evaluated" builtin is available.

sthalik: Alright. Seems like my original proposal wasn't sufficient for what we're trying to achieve. It's not too useful for any of `CORRADE_CONSTEXPR{17,20,23}` to be defined -- C++14 relaxed constexpr is already enough for anything Vector has to handle. But `CORRADE_CONSTEXPR14` needs to reflect whether it works in practice (`__cpp_constexpr` is accurate here). Then we have several kinds of constexpr functions:

- C++11 semantics -- `constexpr`
- constexpr with C++14 rules, no need for inline asm or intrinsics -- `CORRADE_CONSTEXPR14`
- ones that need SIMD -- `CORRADE_CONSTEVAL14`, #ifndef `CORRADE_CONSTEVAL`

The last case can be used like that:

```c++
template<typename T>
CORRADE_CONSTEVAL14 inline T foo(T x)
{
#ifdef CORRADE_CONSTEVAL
    if CORRADE_CONSTEVAL
    {
        // do something traditional to 'x'
    }
    else
#endif
    {
        // do something funny to 'x'
    }
}
```

If you want to enable constexpr when lacking `CORRADE_CONSTEVAL` but when SIMD has been disabled, then it could be set to `constexpr(true)` as long as `__cpp_if_constexpr >= 201606` holds (aka C++17).

sthalik: I think it's better to remove `CORRADE_CONSTEVAL14` from this pull request and put it in Magnum with a name such as `MAGNUM_VECTOR_CONSTEVAL14`. It's too specialized for a global definition. Here it is for reference:

```c++
#ifdef CORRADE_CONSTEVAL
#define CORRADE_CONSTEVAL14 CORRADE_CONSTEXPR14
#else
#define CORRADE_CONSTEVAL14
#endif
```

src/Corrade/Utility/Macros.h
```diff
@@ -394,12 +394,19 @@ Expands to @cpp constexpr @ce on C++14 and newer, empty on C++11. Useful for
 selectively marking functions that make use of C++14 relaxed constexpr rules.
```
mosra, line 394: Since the macro is quite a bit more complex now, I'm thinking a test could be a good idea -- there's `src/Corrade/Utility/MacrosCppXYTest.cpp` which currently test C++17 and 20 features, so a new `MacrosCpp14Test.cpp` (built [only on C++14-capable compilers](https://github.com/mosra/corrade/blob/4ff543dc3f7372f5a05c9d69193e2d37bf328bae/src/Corrade/Test/CMakeLists.txt#L61-L72)) would contain a test for this macro:

```cpp
...

constexpr int sumInAStupidWay(int number) {
    int sum = 0;
    for(int i = 0; i != number; ++i)
        sum += i;
    return sum;
}

void MacrosCpp14Test::constexpr14() {
    #ifndef CORRADE_CONSTEXPR14
    CORRADE_SKIP("CORRADE_CONSTEXPR14 not available on this compiler.");
    #else
    CORRADE_COMPARE(sumInAStupidWay(17), 153);
    #endif
}

}}}}

CORRADE_TEST_MAIN(Corrade::Utility::Test::MacrosCpp14Test)
```

Because without the test it's hard to verify that the macro is actually correctly defined only where C++14 constexpr works.

And the docs should be extended to mention the compiler versions on which the macro is defined.

src/Corrade/Utility/Macros.h
```diff
@@ -394,12 +394,19 @@ Expands to @cpp constexpr @ce on C++14 and newer, empty on C++11. Useful for
...
 @see @ref CORRADE_CXX_STANDARD
 */
-#if CORRADE_CXX_STANDARD >= 201402
+#if defined __cpp_constexpr && __cpp_constexpr >= 201304 || \
+    defined CORRADE_TARGET_MSVC && _MSC_VER >= 1910 && CORRADE_CXX_STANDARD >= 201402L
```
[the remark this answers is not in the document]
sthalik: > Shouldn't this contain a bunch of parentheses, to ensure && and || are evaluated in the correct order?

It always binds tightest in the order: `>=`, `&&`, `||`. I'm going to add the parentheses anyway.

> What's the L needed for? I don't think it's essential.

Good to know.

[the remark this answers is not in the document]
mosra: > I'm going to add the parentheses anyway.

Thank you :) I can never remember the order, and GCC sometimes warns about "|| within &&" and such if I forget the braces, so I suspect it's always different that I would expect :D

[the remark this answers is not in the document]
sthalik: This table always helps: https://en.cppreference.com/w/cpp/language/operator_precedence.

src/Corrade/Utility/Macros.h
sthalik: Generally when testing constexpr functions I recommend putting the compiler in a situation where it can't weasel out of performing the computation at compile-time. Hence the static assert.

mosra: Ah yes, good idea. Except that now you're using the C++17 variant of it ;)

src/Corrade/Utility/Macros.h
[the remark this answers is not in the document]
mosra: Wait so the second MSVC-specific part is no longer needed? I don't see it in the last revision.

src/Corrade/Utility/Macros.h
sthalik: I think I managed to test whether the function is constexpr through SFINAE.

src/Corrade/Utility/Test/MacrosCpp14Test.cpp
```diff
@@ -0,0 +1,65 @@
...
+        CORRADE_SKIP("CORRADE_CONSTEXPR14 not available on this compiler.");
+    else {
+        CORRADE_CONSTEXPR14 int sum = sumInAStupidWay(17);
+        CORRADE_COMPARE(sum, 136);
+    }
+}
+
+MacrosCpp14Test::MacrosCpp14Test() {
+    addTests({&MacrosCpp14Test::constexpr14});
+}
```
mosra, line 61: Can you follow the same method order as in other tests? Constructor first, helper functions defined next to where they get used, etc.

src/Corrade/Utility/Test/MacrosCpp14Test.cpp
```diff
@@ -0,0 +1,65 @@
...
+    int sum = 0;
+    for(int i = 0; i != number; ++i)
+        sum += i;
+    return sum;
+}
+
+template<int> struct sfinae : std::true_type{};
+template<class T> sfinae<(T::sumInAStupidWay(0), 0)> check(int);
+template<class> std::false_type check(...);
+template<class T> struct hasRelaxedConstexpr : decltype(check<T>(0)){};
```
mosra, line 48: But this way you're not testing if the macro is actually defined to `constexpr` when it should be, no? Like, the test would pass even if `CORRADE_CONSTEXPR14` would be empty always, or if you wouldn't use it at all.

... and you're not using the static assert anymore :P

There really needs to be something based on whether the macro is empty or not.

mosra (commented):
:eyes:

src/Corrade/Utility/Test/MacrosCpp14Test.cpp
sthalik: 🆗

src/Corrade/Utility/Test/MacrosCpp14Test.cpp
sthalik: Alright, so I added an `#if 0` so that it can be verified. Sadly static assert broke the build.

src/Corrade/Utility/Macros.h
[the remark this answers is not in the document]
sthalik: Good catch! It was only in the Magnum PR.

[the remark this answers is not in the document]
mosra: This is still missing the parentheses...

src/Corrade/Utility/Macros.h
mosra: Still needs the documentation update...

src/Corrade/Utility/Test/MacrosCpp14Test.cpp
mosra: I don't like this solution, to be honest. Because it requires manual fiddling with the `#if 0`, it's impossible to test on the CI, and thus it's the same as if the whole test file wouldn't be here.

sthalik: What then, you want the test to fail rather than skip?

mosra: The test would pass even if `CORRADE_CONSTEXPR14` wasn't used anywhere in it, that's the problem I have with it. It's not testing the thing it should be testing, so it's useless.

sthalik: It'll skip on and old compiler, that's why that sfinae logic is there in the first place. Toggle the `#if 0` and see.

mosra:
Actually, I'm still not sure what is the main reason the change to `CORRADE_CONSTEXPR14` is needed. 

Apart from MSVC, are there *any* compilers that report C++14 as supported via the `__cplusplus` macro but don't actually support C++14 constexpr? I suspect not, or at least no relevant versions that'd be widely used in distros or OSes. And since you have another dedicated check for MSVC >= 2017 in the `#if` (does that mean `__cpp_constexpr` didn't work?), I think the macro could still check for `CORRADE_CXX_STANDARD >= 201402` like before. Which *significantly* simplifies everything, since the test doesn't need to handle the case of "C++14 supported but C++14 constexpr not" at all.

Then, for MSVC, isn't it just about excluding MSVC 2015? It would boil down to 

```cpp
#if CORRADE_CXX_STANDARD >= 201402 && !defined(CORRADE_MSVC2015_COMPATIBILITY)
#define CORRADE_CONSTEXPR14 constexpr
...
#endif

**Comment by sthalik** [comment]
> Apart from MSVC, are there any relevant compilers that report C++14 as supported via the __cplusplus macro but don't actually support C++14 constexpr (is that scenario even allowed by the standard?)?

GCC 4.8 doesn't implement relaxed constexpr either.

> and using CORRADE_MSVC2015_COMPATIBILITY to special-case the test, just for this one compiler, assuming sanity everywhere else

Then what about Magnum's vector code?

**Comment by sthalik** [comment]
> Apart from MSVC, are there any relevant compilers that report C++14 as supported via the __cplusplus macro but don't actually support C++14 constexpr

Compilers tend to expose `-std=c++14` before it's fully implemented or even before the standard is finalized. Quite a few of the CI bots are running GCC 4.8 which has this problem.

https://gcc.gnu.org/projects/cxx-status.html -- 4.3 -> 5.0*
https://clang.llvm.org/cxx_status.html - 2.9 -> 3.4*
https://learn.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance - 2013 12.0 -> 2017 15.0*

> Which significantly simplifies everything, since the test doesn't need to handle the case of "C++14 supported but C++14 constexpr not" at all.

The whole point is to handle that for cases such as Magnum vector code. Or am I meant to litter it with `_MSC_VER` and `__GNUG__`?

**Comment by mosra** [comment]
> Compilers tend to expose `-std=c++14` before it's fully implemented

Yes, `-std=c++xy` is, but I'm talking about the `__cplusplus` macro, which is defined to `2014xy` only on GCC 5+. Which solves the problem for GCC 4.8, it was already working there correctly even before this PR. Old Clangs are irrelevant nowadays, so there's no point in distinguishing anything there; and for MSVC we have to special-case anyway.

I'm not trying to solve the general case of "is feature X supported completely and without bugs on compiler Y", here it's just about having the simplest possible condition for "is C++14 constexpr supported on compilers we care about", and since C++14 is rather old and the relevant compilers are dying out, it can be coarser than, say, checking for `consteval` support. Which is what `#if CORRADE_CXX_STANDARD >= 201402 && !defined(CORRADE_MSVC2015_COMPATIBILITY)` does.

> The whole point is to handle that for cases such as Magnum vector code. Or am I meant to litter it with `_MSC_VER` and `__GNUG__`?

What? Why? The whole point here is to define the macro in a way that does the right thing without a need for extra checks every time it's used, and have a test that verifies it indeed does the right thing.

- - - 

I guess easiest is if I just do the thing instead of talking about it. The `CORRADE_CONSTEXPR14` part is merged as fbb9305265fd2fe89e45893108f9e0ee686c341d, with the corresponding test in bee514f31d1d58e0905683dab3d23a903ff930a8, for both the C++11 and the C++14 case. All CIs pass.

File: src/Corrade/Utility/Macros.h [hunk 6]
```diff
@@ -390,16 +390,38 @@ regarding RAII.
...
+@m_since{2022,10}
+
+Expands to a predicate determining whether given @cpp constexpr @ce function
+is being evaluated at compile-time or not. Under C++14 rules, constexpr
+functions may be defined as long as at least some of the code paths are able
+to be executed at compile-time. As long as a fallback is present, features
+such as SIMD or inline assembly may be used in functions marked constexpr.
+
+This support is available on all C++20 capable compilers, but also certain
+ones (Clang 9, GCC 9, MSVC 17.1) that expose the feature as a non-portable
```
mosra, line 416:
```suggestion
ones (Clang 9, GCC 9, MSVC 2022 17.1) that expose the feature as a non-portable
```

src/Corrade/Utility/Macros.h
```diff
@@ -390,16 +390,38 @@ regarding RAII.
...
+extension. In which case it may be used under C++14 relaxed constexpr rules.
+*/
+#if (defined(__clang__) && __clang_major__ >= 9) || (defined(__GNUG__) && __GNUG__ >= 9) || (defined(_MSC_VER) && _MSC_VER >= 1931)
```
mosra, line 419:
```suggestion
#if (defined(CORRADE_TARGET_CLANG) && !defined(CORRADE_TARGET_APPLE_CLANG) && __clang_major__ >= 9) || (defined(CORRADE_TARGET_APPLE_CLANG) && __clang_major__*100 + __clang_minor__ >= 1104) || (defined(CORRADE_TARGET_GCC) && __GNUC__ >= 9) || (defined(CORRADE_TARGET_MSVC) && _MSC_VER >= 1931)
```

This handles also Apple Clang. For clang-cl I checked the CIs and even MSVC 2019 seems to bundle Clang 12, so the check for `_MSC_VER` should suffice.

src/Corrade/Utility/Macros.h
```diff
@@ -390,16 +390,38 @@ regarding RAII.
...
+#define CORRADE_CONSTEVAL (__builtin_is_constant_evaluated())
+#elif CORRADE_CXX_STANDARD >= 202002
+#define CORRADE_CONSTEVAL (std::is_constant_evaluated())
```
mosra, line 422: Since this depends on `<type_traits>`, I'm thinking the macro should go to `TypeTraits.h` instead. The `Macros.h` header deliberately doesn't pull in any STL headers.

src/Corrade/Utility/Macros.h
```diff
@@ -390,16 +390,38 @@ regarding RAII.
...
+functions may be defined as long as at least some of the code paths are able
+to be executed at compile-time. As long as a fallback is present, features
+such as SIMD or inline assembly may be used in functions marked constexpr.
+
+This support is available on all C++20 capable compilers, but also certain
+ones (Clang 9, GCC 9, MSVC 17.1) that expose the feature as a non-portable
+extension. In which case it may be used under C++14 relaxed constexpr rules.
+*/
+#if (defined(__clang__) && __clang_major__ >= 9) || (defined(__GNUG__) && __GNUG__ >= 9) || (defined(_MSC_VER) && _MSC_VER >= 1931)
+#define CORRADE_CONSTEVAL (__builtin_is_constant_evaluated())
```
mosra, line 420: I'm thinking `CORRADE_IS_CONSTANT_EVALUATED` would be a less confusing name.

src/Corrade/Utility/Macros.h
```diff
@@ -390,16 +390,38 @@ regarding RAII.
...
+Expands to a predicate determining whether given @cpp constexpr @ce function
+is being evaluated at compile-time or not. Under C++14 rules, constexpr
+functions may be defined as long as at least some of the code paths are able
+to be executed at compile-time. As long as a fallback is present, features
+such as SIMD or inline assembly may be used in functions marked constexpr.
+
+This support is available on all C++20 capable compilers, but also certain
+ones (Clang 9, GCC 9, MSVC 17.1) that expose the feature as a non-portable
+extension. In which case it may be used under C++14 relaxed constexpr rules.
+*/
```
mosra, line 418: Thanks for the docs, now it's just a test missing :) `TypeTraitsCpp20Test.cpp`

sthalik:
Note, it's specifically *not* meant to be used as `if constexpr(std::is_constant_evaluated) ...`, because in that case it always returns true.

Issue #123, opened 2021-10-07
constexpr StringView without literals


jdumas:
Hi. I noticed that a `constexpr Containers::StringView` can only be created from `const char *` via a literal `_s`. Reading the doc this seems to be a deliberate design decision, but I'm not entirely sure why? E.g., the tag `Global` is documented as:

> The referenced string is global, i.e., with an unlimited lifetime. Enabling this flag can avoid needless allocations and copies in cases where the consumer needs to extend lifetime of passed string view.

But why would you want to extend the lifetime of a passed string view? This is a view, not a smart pointer no?

mosra:
A `constexpr` creation of a `StringView` from an `ArrayView` mentioned above got implemented in 81b708905069121b056f41fed37a6ec08fc12b12, so the following is now possible, at least, although I'm not sure if it's any better than writing a helper that consumes `char(&)[n]` and passes that to the already-`constexpr` StringView constructor taking a pointer + size:

```cpp
constexpr Containers::StringView a = Containers::arrayView("hello").exceptSuffix(1);
```

Everything else, related to `constexpr`-aware construction from a `char*`, needs something like #152, and even then I'm not sure if it's a good idea given that it needs something like a compile-time `while` loop that iterates over the array until the first `'\0'` is found to be *really* compatible with what `strlen()` does. Worth investigating is also https://gcc.gnu.org/bugzilla/show_bug.cgi?id=86590 and https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80265 where the GCC devs deal with poor codegen related to doing this very operation.

PR #155, opened 2022-11-10
Containers: add StringView overload for known sizes

sthalik:
Is this a reasonable compromise between support for unterminated strings and the need for a constexpr constructor?


sthalik:
Actually this isn't a good idea. I forgot about you mentioning `operator""_s`.

mosra:
I plan to have a `constexpr` constructor relying on compiler-specific magic, such as the one that's used to [implement `constexpr` construction of `std::string_view`](https://github.com/mosra/corrade/issues/123#issuecomment-937903431). But, due to the compiler-specific magic, only as an optional "nice to have" optimization that removes the need for runtime `strlen()` even when `_s` literals are not used, not something people should rely on.

For constexpr creation from arrays I plan to update the `ArrayView` construxctor to be `constexpr`, as detailed in #123, when I figure out how without adding a header dependency.

PR #158, opened 2022-11-13
PluginManager: fix static test build

sthalik:
This was preventing cmake configuring from finishing.

### DIFF
--- a/src/Corrade/PluginManager/Test/animals/CMakeLists.txt
+++ b/src/Corrade/PluginManager/Test/animals/CMakeLists.txt
@@ -73,6 +73,6 @@ if((CORRADE_BUILD_STATIC OR CORRADE_TARGET_WINDOWS) AND NOT CORRADE_PLUGINMANAGE
 endif()
 
 # Additionally the plugin dependencies have to be linked as well
-if(CORRADE_TARGET_WINDOWS)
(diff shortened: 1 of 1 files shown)


mosra:
Huh? `CORRADE_TARGET_WINDOWS` and `CORRADE_PLUGINMANAGER_NO_DYNAMIC_PLUGIN_SUPPORT` are mutually exclusive, there's never a case when both would be set.

sthalik:
Why not? It's working on my end with no further modifications with clang/lld, gcc/binutils and msvc.

sthalik:
If there are any edge cases in some of the plugins that I'm not using (only stb and tga importers right now), I might be able to fix them on win32 and send a pull request for that. Then the option doesn't have to remain exclusive with win32.

The goal is to have a monolithic binary.

mosra:
There's never a case when both would be set. The variable is set for platforms that don't have a (practically usable) concept of a dynamically loadable library, such as Emscripten or Android.

Or are you setting `CORRADE_PLUGINMANAGER_NO_DYNAMIC_PLUGIN_SUPPORT` yourself? :scream:

sthalik:
I'm using it. The goal is to remove any possible calls to stale plugins that might exist. And also remove the message on stderr about plugin search paths.

mosra:
That's not a supported use case, sorry. I'm not interested in the extra burden of maintaining that.

Setting the plugin directory to, say, `"nonexistent"`, in the constructor, is the supported way currently. There might eventually be an explicit option to disable static plugins, but so far I'm not planning that.

Issue #157, opened 2022-11-12
std::tuple_size / tuple_element specializations for Corrade containers


mosra:
Revived from #127. For C++17 `auto [a, b] = someType;`, among others:

- [ ] `Containers::Pair`, for which the `std::tuple_whatever` specializations could go into the already-existing `Containers/PairStl.h`, and a corresponding new C++17-only `PairStlCpp17Test.cpp`
- [ ] `Containers::Triple`, for which the `std::tuple_whatever` specializations could go into the already-existing `Containers/TripleStl.h`, and a corresponding new C++17-only `TripleStlCpp17Test.cpp`
- [ ] `Containers::StaticArrayView`, for which it could go into the already-existing `Containers/ArrayViewStl.h`, and a corresponding new `ArrayViewStlCpp17Test.cpp`
- [ ] `Containers::StaticArray`, into a new `StaticArrayStl.h` header, and a new `StaticArrayStlCpp17Test.cpp` as well

The assumption is that there will be more and more C++14/17/20-specific features, so the ton of new files is fine.

The `std::tuple_element` / `std::tuple_size` is a C++11 feature already so I wouldn't hide them in any `#if compiling_as_cpp17`. Having them exposed unconditionally would mean `std::get<N>()` working with Corrade/Magnum types even under C++11, for example.

- [ ] Maybe it would make sense to test them directly in C++11 test files somehow, instead of C++17-only tests that won't be run on half the CI jobs. Like, for example, just trying to call `std::get<N>()` on them, and if that works, the C++17 structured bindings should work too.

And for Magnum a similar treatment would be done for `Math::Vector` and possibly other math types, in a similar way.

mosra:
Done in 5f4fc52ad2f07c4e022e8ba8e770144becb69b17, together with structured bindings for `StridedDimensions`.

PR #159, opened 2022-11-13
Containers: add structured bindings for Pair

sthalik:
Somehow `decltype()` of a struture binding doesn't have a reference. So test it at runtime using the `auto& [e1, e2]` case.


src/Corrade/Containers/PairStl.h
```diff
@@ -41,7 +41,22 @@ construction and assignment. See @ref Containers-Pair-stl for more information.
...
+template<class F, class S> struct tuple_element<0, Corrade::Containers::Pair<F, S>> { using type = F; };
+template<class F, class S> struct tuple_element<1, Corrade::Containers::Pair<F, S>> { using type = S; };
+}
+
+namespace Corrade { namespace Containers {
+
+#if CORRADE_CXX_STANDARD >= 201703
+template<std::size_t N, class F, class S> constexpr const std::tuple_element_t<N, Pair<F, S>> & get(const Pair<F, S>& value) { if constexpr(N == 1) return value.second(); return value.first(); }
+template<std::size_t N, class F, class S> constexpr std::tuple_element_t<N, Pair<F, S>>& get(Pair<F, S>& value) { if constexpr(N == 1) return value.second(); return value.first(); }
+template<std::size_t N, class F, class S> constexpr std::tuple_element_t<N, Pair<F, S>> get(Pair<F, S>&& value) { if constexpr(N == 1) return value.second(); return value.first(); }
```
mosra, line 56: Hmm, now I checked, and there isn't a way around polluting my namespaces with `get()`, right? Sigh.

Somehow I thought `std::tuple_element` specialization would *itself* contain the `get()`, because that would make total sense, and would be quite clean, but apparently not?

That's also why I thought this could be enabled for C++11 too, to make `std::get<N>()` working on these types... but if not, then it just doesn't make sense to have this there. And I don't see any other use cases to have this on C++11, so having it C++17-only is fine.

I need to think a bit whether go with this namespace-polluting way, or whether just go with `get` as a member. I'll get back to you on that.

mosra, line 56: Didn't you mean to return a `&&`? (Apologies if I'm missing something obvious.)

```suggestion
template<std::size_t N, class F, class S> constexpr std::tuple_element_t<N, Pair<F, S>>&& get(Pair<F, S>&& value) { if constexpr(N == 1) return value.second(); return value.first(); }
```

src/Corrade/Containers/PairStl.h
```diff
@@ -41,7 +41,22 @@ construction and assignment. See @ref Containers-Pair-stl for more information.
...
+namespace std {
+template<class F, class S> struct tuple_size<Corrade::Containers::Pair<F, S>> : std::integral_constant<std::size_t, 2> {};
+template<class F, class S> struct tuple_element<0, Corrade::Containers::Pair<F, S>> { using type = F; };
+template<class F, class S> struct tuple_element<1, Corrade::Containers::Pair<F, S>> { using type = S; };
+}
+
+namespace Corrade { namespace Containers {
+
+#if CORRADE_CXX_STANDARD >= 201703
+template<std::size_t N, class F, class S> constexpr const std::tuple_element_t<N, Pair<F, S>> & get(const Pair<F, S>& value) { if constexpr(N == 1) return value.second(); return value.first(); }
```
mosra, line 54: Given this is C++17, would it work with just `auto` (explicit `const auto&`, so it doesn't decay too much)? Or is that not a valid I'm trying to see how it could work directly on the Pair class without having to rely on `std::tuple_element`.

```suggestion
template<std::size_t N, class F, class S> constexpr const auto& get(const Pair<F, S>& value) {
    if constexpr(N == 1) 
        return value.second(); 
    return value.first();
}
```

Also wrapping wouldn't hurt :P 

Same for others.

src/Corrade/Containers/Test/PairStlTest.cpp
```diff
@@ -83,6 +85,46 @@ void PairStlTest::convertMove() {
     CORRADE_VERIFY(!c.second);
 }
 
+void PairStlTest::structuredBindings()
+{
+#if CORRADE_CXX_STANDARD >= 201703
```
mosra, line 90: This way it won't get tested on *any* CI jobs, and neither for anyone who doesn't explicit explicitly pass `-std=c++17` to the whole project. Which people rarely do.

It needs an explicit new test file, as I mentioned in #127 (and #157).

mosra:
Just for the record, a variant of this, together with structured bindings for `StaticArray[View]` and `StridedDimensions` was commited in 5f4fc52ad2f07c4e022e8ba8e770144becb69b17.

PR #161, opened 2022-12-09
modules: fix typos.

pezcode:
We recently switched to spell-checking our code, and found these in the copied modules/ files. Figured I might as well fix it upstream 🍌


mosra:
Merged as 469c501d68ba2d3f3644ac80d0bf1acc99985dd6, thank you!

PR #165, opened 2023-01-19
debuggers: adapt MSVC natvis to recent String changes

pezcode:
Small fix to make the .natvis file work with current master again 🍒 And I made MSVC show you whether a string is small-size-optimized or not, sometimes handy to know.


mosra:
Thank you!

Which reminds me, the GDB pretty-printers should probably get an update as well -- @williamjcm, is that something you'd have time for? Thanks as well!

williamjcm:
I do have time for it.

Out of curiosity, were there any other changes in Corrade that could affect the printers ?

mosra:
Thank you!

There's a new `Iterable` and `StringIterable` containers that got added since you made the original PR in August, and I'll add a `StridedBitArrayView` (and possibly `StaticBitArray`) soon to have a bit-sized complement for all the most important `Array[View]` types. Apart from that, the change to `StringView` internals was the only that would be breaking anything, I don't remember anything else.

PR #184, opened 2024-09-22
Fix ambiguous use of format with new Apple Clang and C++20

Twinklebear:
I just updated XCode which updated to `AppleClang 16.0.0.16000026` and the found the unqualified uses of `format` here now produce a compile error about the call to `format` being ambiguous. This PR fixes the compile issue by qualifying the use with the `Utility` namespace. The ambiguous use occurs when targetting C++20, as I didn't repro it when just building corrade on its own, but in a project that enables C++20 I do see the ambiguous use error. If I target C++20 when building corrade, I can repro the compile issue:

```
In file included from /Users/will/repos/corrade/src/Corrade/Utility/rc.cpp:30:
/Users/will/repos/corrade/src/Corrade/Utility/Implementation/ResourceCompile.h:217:12: error: call to 'format' is ambiguous
  217 |     return format(R"(/* Compiled resource file. DO NOT EDIT! */
      |            ^~~~~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX15.0.sdk/usr/include/c++/v1/__format/format_functions.h:476:1: note: candidate function [with _Args = <std::string &, std::string &, Corrade::Containers::BasicStringView<const char>, Corrade::Containers::String, std::string &, const Corrade::Containers::BasicStringView<const char> &, const Corrade::Containers::BasicStringView<const char> &, unsigned long, const char *>]
  476 | format(format_string<_Args...> __fmt, _Args&&... __args) {
      | ^
/Users/will/repos/corrade/src/Corrade/Utility/Format.h:349:71: note: candidate function [with Args = <std::string, std::string, Corrade::Containers::BasicStringView<const char>, Corrade::Containers::String, std::string, Corrade::Containers::BasicStringView<const char>, Corrade::Containers::BasicStringView<const char>, unsigned long, const char *>, String = Corrade::Containers::String, MutableStringView = Corrade::Containers::BasicStringView<char>]
  349 | template<class ...Args, class String, class MutableStringView> String format(const char* format, const Args&... args) {
      |                                                                       ^
```

### DIFF
--- a/src/Corrade/Utility/Implementation/ResourceCompile.h
+++ b/src/Corrade/Utility/Implementation/ResourceCompile.h
@@ -115,7 +115,7 @@ Containers::String resourceCompile(const Containers::StringView name, const Cont
 
     /* Special case for empty file list */
     if(files.isEmpty()) {
-        return format(R"(/* Compiled resource file. DO NOT EDIT! */
+        return Utility::format(R"(/* Compiled resource file. DO NOT EDIT! */
 
 #include "Corrade/Corrade.h"
 #include "Corrade/Utility/Macros.h"
@@ -214,7 +214,7 @@ int resourceFinalizer_{0}() {{
        about functions which don't have corresponding declarations (enabled by
        -Wmissing-declarations in GCC). If we don't have any data, we don't
        create the resourceData array, as zero-length arrays are not allowed. */
-    return format(R"(/* Compiled resource file. DO NOT EDIT! */
+    return Utility::format(R"(/* Compiled resource file. DO NOT EDIT! */
 
 #include "Corrade/Corrade.h"
 #include "Corrade/Utility/Macros.h"
@@ -361,7 +361,7 @@ Containers::String resourceCompileSingle(const Containers::StringView name, cons
         dataHexcode.resize(dataHexcode.size() - 2);
     }
 
-    return format(R"(/* Compiled resource file. DO NOT EDIT! */
+    return Utility::format(R"(/* Compiled resource file. DO NOT EDIT! */
 
 extern const unsigned int resourceSize_{0} = {1};
 extern const unsigned char resourceData_{0}[] = {{


falbrechtskirchinger:
Was just about to submit the same. Has nothing to do with Apple Clang specifically. libc++ implements C++20's `std::format` now, hence the ambiguity. Seeing this with libstdc++ (using various recent versions of GCC/Clang) as well.

mosra:
Thank you!

@falbrechtskirchinger yeah I saw the same change on your fork some weeks ago and wanted to investigate further, but then it somehow fell through the cracks -- sorry. (And earlier on @sthalik's fork too, but at that time it indeed felt like a MSVC 2022 bug rather than something else.)

I suppose this is due to ADL, where if it's called as `format(..., <some STL type>)`, the compiler then can choose between the implementation inside `Corrade::Utility` and the one inside `std`, if some STL header happens to transitively include `<format>` as well (which seems to be the case on Clang + libc++ and on MSVC, but not on GCC's libstdc++ for example). This is also why it happens only when compiling `rc.cpp`, because that code is one of the last areas that still uses `std::string`. And same would I guess happen if `std::format()` would be called unqualified with a type from `Corrade::Utility`.

And similar problems are with any new likely-to-conflict names introduced in the `std` namespace, like `empty()`, `lerp()` or `size()`... Yay this language feature :see_no_evil:

Pushed a slightly different variant of the fix as cff4f2ee6f871c8cf4630bf6928fd3ea71734653. This fixed the issue on my end (tho only if I explicitly `#include <format>` in that file), should do the job on your end too.

Twinklebear:
That fixes it for me, thanks!

falbrechtskirchinger:
> @falbrechtskirchinger yeah I saw the same change on your fork some weeks ago and wanted to investigate further, but then it somehow fell through the cracks -- sorry. (And earlier on @sthalik's fork too, but at that time it indeed felt like a MSVC 2022 bug rather than something else.)

No worries. I'll be able to verify once I'm done with the non-graphics portion of my code. If there's still an issue, I'll be sure to let you know. Thanks!

PR #186, opened 2024-10-05
updating rpm spec, adding rpm build script.

alkavan:
Added build script for rpm package and updated `.spec` file.


package/rpm/README.md
```diff
@@ -0,0 +1,11 @@
+# Corrade RPM Package
+
+You will need to install the following dependencies for build rpms:
+```
+sudo dnf install fedora-packager rpmdevtools gcc
+```
+
+Create user directory tree building RPMs:
+```
+mkdir -p ~/rpmbuild/{BUILD,BUILDROOT,RPMS,SOURCES,SPECS,SRPMS}
```
mosra, line 10: #125 calls `rpmdev-setuptree`, would that be the equivalent to this?

I suppose it could be called inside `build.sh` directly.

package/rpm/build.sh
```diff
@@ -0,0 +1,23 @@
...
+echo "** repository hash: ${version_hash} ..."
+
+# archive reository
+(cd ../.. && git archive --format=tar.gz --prefix=corrade-${version_hash}/ -o ~/rpmbuild/SOURCES/corrade-${version_hash}.tar.gz HEAD)
+echo "** created archive: ~/rpmbuild/SOURCES/corrade-${version_hash}.tar.gz"
+sleep 2
+
+# replace spec version
+spec_file="corrade.spec"
+cp ${spec_file} ${spec_file}.build
```
mosra, line 14: I'm actually fine with the `corrade.spec` file getting overwritten by `sed` in-place -- it's all tracked by Git anyway, so there's no chance of accidental information loss.

ArchLinux does the same to the `PKGBUILD` files when building them and it has a nice side effect of knowing for what revision was the package script last updated. (And yes, just commit the `*.spec` file with that updated version hash.)

mosra:
Moved the README contents to the documentation, did some minor adjustments and merged as 9ef83b7291371900fd24c307bac66433941f2114. I hope I didn't make any obvious errors that prevent this from working, haha.

PR #192, opened 2024-11-21
Fix cast precedence order change on emscripten 3.1.65+

Twinklebear:
I started noticing a bad load/crash on Emscripten 3.1.65+ in our app (see https://github.com/emscripten-core/emscripten/issues/22794 ). It turns out that the issue is that the casting precedence for `size_t * ptrdiff_t` is undefined/decidable by the compiler/etc, and in the past (<= 3.1.64) Emscripten must have ended up casting to a signed type but on 3.1.65+ it ends up casting the ptrdiff to an unsigned type. When using the GLTF importer plugin (or anything that uses `flipped`) a negative stride is used to iterate through the view in reverse. However, when the negative stride is cast to unsigned in the multiplies this PR changes, it overflows and causes an out of bounds access.

This PR fixes (at least a few of) these locations where we have a `size_t` multiplied by a stride `ptrdiff_t`, where a negative stride would cause some overflow to occur. With these changes (actually we just need the change on line 2219, I updated a few other spots that seemed like they would need it), our app works on Emscripten 3.1.65+

### DIFF
--- a/src/Corrade/Containers/StridedArrayView.h
+++ b/src/Corrade/Containers/StridedArrayView.h
@@ -2216,12 +2216,12 @@ namespace Implementation {
             return StridedArrayView<dimensions - 1, T>{
                 Size<dimensions - 1>(size._data + 1, typename Implementation::GenerateSequence<dimensions - 1>::Type{}),
                 Stride<dimensions - 1>(stride._data + 1, typename Implementation::GenerateSequence<dimensions - 1>::Type{}),
-                static_cast<typename std::conditional<std::is_const<T>::value, const char, char>::type*>(data) + i*stride._data[0]};
+                static_cast<typename std::conditional<std::is_const<T>::value, const char, char>::type*>(data) + static_cast<std::ptrdiff_t>(i)*stride._data[0]};
         }
     };
     template<class T> struct StridedElement<1, T> {
         static T& get(typename std::conditional<std::is_const<T>::value, const void, void>::type* data, const Size1D&, const Stride1D& stride, std::size_t i) {
-            return *reinterpret_cast<T*>(static_cast<typename std::conditional<std::is_const<T>::value, const char, char>::type*>(data) + i*stride._data[0]);
+            return *reinterpret_cast<T*>(static_cast<typename std::conditional<std::is_const<T>::value, const char, char>::type*>(data) + static_cast<std::ptrdiff_t>(i)*stride._data[0]);
         }
     };
 }
@@ -2246,7 +2246,7 @@ template<unsigned dimensions, class T> T& StridedArrayView<dimensions, T>::opera
     for(std::size_t j = 0; j != dimensions; ++j) {
         CORRADE_DEBUG_ASSERT(i._data[j] < _size._data[j],
             "Containers::StridedArrayView::operator[](): index" << i << "out of range for" << _size << "elements", *reinterpret_cast<T*>(static_cast<ArithmeticType*>(_data)));
-        data += i._data[j]*_stride._data[j];
+        data += static_cast<std::ptrdiff_t>(i._data[j])*_stride._data[j];
     }
 
     return *reinterpret_cast<T*>(data);
@@ -2261,7 +2261,7 @@ template<unsigned dimensions, class T> StridedArrayView<dimensions, T> StridedAr
     Containers::Size<dimensions> size = _size;
     size._data[0] = std::size_t(end - begin);
     return StridedArrayView<dimensions, T>{size, _stride,
-        static_cast<ArithmeticType*>(_data) + begin*_stride._data[0]};
+        static_cast<ArithmeticType*>(_data) + static_cast<std::ptrdiff_t>(begin)*_stride._data[0]};
 }
 
 template<unsigned dimensions, class T> template<unsigned newDimensions> StridedArrayView<newDimensions, T> StridedArrayView<dimensions, T>::slice(const Containers::Size<dimensions>& begin, const Containers::Size<dimensions>& end) const {


Twinklebear:
I'm not sure on the failure of the `linux-threadsanitizer` test:

```
ninja $NINJA_JOBS
[253/515] Compiling static plugin metadata file /root/project/build/src/Corrade/PluginManager/Test/init-fini/resource_InitFiniStatic.cpp
FAILED: src/Corrade/PluginManager/Test/init-fini/resource_InitFiniStatic.cpp 
cd /root/project/src/Corrade/PluginManager/Test/init-fini && /root/project/build/Debug/bin/corrade-rc InitFiniStatic --single InitFiniStatic.conf /root/project/build/src/Corrade/PluginManager/Test/init-fini/resource_InitFiniStatic.cpp
FATAL: ThreadSanitizer: unexpected memory mapping 0x768ba6a00000-0x768ba6d52000
[255/515] Building CXX object src/Corrade/Utility/CMakeFiles/CorradeUtilityTestLib.dir/Arguments.cpp.o
ninja: build stopped: subcommand failed.
```

I don't know if I have ability to rerun on circle CI, I can try to run it on a linux machine I have if it should be passing

mosra:
Huh, this sounds really bad :eyes: Thanks for doing all the investigation for this, much appreciated.

I think I ran into something similar in the past elsewhere, where depending on machine bitness a multiplication of different types resulted in something really strange. If only there was a way to disable C's implicit conversions everywhere, to force the code to explicitly cast, sigh... Because I fear that even if this fixes most _current_ occurences, it'll inevitably slip through again in future code :/

I'll add regression test cases for all the spots you fixed and merge this.

- - -

The threadsanitizer failure is expected, see #191 for details. Unfortunately I ran out of ideas how to fix it, so it's going to keep failing for a while :/

Twinklebear:
I just wanted to check in, did you end up fixing this in master separately or would merge this PR to fix the spot I changed + fix the rest? I was hesitant to modify a larger amount of the code, since I'm not familiar.

mosra:
Sorry for the delay, I was just too busy last week. I'll try to merge your change + add regression tests tomorrow.

Twinklebear:
No worries! Thanks for your work on magnum and corrade

mosra:
Ah well. I really _really_ wanted to add a regression test, but I just cannot reproduce in an isolated case. Turns out that just this

```patch
diff --git a/src/MagnumPlugins/GltfImporter/GltfImporter.cpp b/src/MagnumPlugins/GltfImporter/GltfImporter.cpp
index ce08e7577..053354794 100644
--- a/src/MagnumPlugins/GltfImporter/GltfImporter.cpp
+++ b/src/MagnumPlugins/GltfImporter/GltfImporter.cpp
@@ -3252,6 +3252,7 @@ namespace {
 template<class T, class F, class G> std::size_t stableSortRemoveDuplicatesToPrefix(Containers::ArrayView<T> container, F lessThanComparator, G equalComparator) {
     std::stable_sort(container.begin(), container.end(), lessThanComparator);
     const auto reversed = stridedArrayView(container).template flipped<0>();
+    !Debug{} << &*(reversed.begin() + 1);
     return std::unique(reversed.begin(), reversed.end(), equalComparator) - reversed.begin();
 }
```

i.e., _printing_ an address, makes the thing compile differently somehow and the issue no longer happens (on 3.1.72, at least). Apart from that, the multiplication produces an overflowing `size_t` even natively on x86, so it doesn't seem like this is due to a change in some cast precedence order, but rather constant folding / optimization happening differently somehow, resulting in the overflow no longer being cancelled out.

In other words, this is what seems to be happening:

```cpp
ptr = i * stride +   data  // for negative stride, `data > i*-stride` always
ptr = 5 *   -16  + 0xd40c4
ptr = 4294967216 + 0xd40c4 // the multiplication overflows, staying as a unsigned 32-bit value
ptr = 0xd4074 // with operands being 32-bit, the addition overflows again, giving correct output
```

As far as my (poor) understanding goes, overflow is well defined for unsigned types, so _technically_ this shouldn't be a problem? I think? On a 64-bit platform the same overflow / wraparound happens, just with much larger values. So my suspicion here is that, somehow, when the particular optimization in 3.1.66+ kicks in, the wraparound is no longer happening, ultimately resulting in the final address being outside of the 32-bit range. When I change the affected calculation to be performed on full 64-bit values like this, then it doesn't happen either:

```patch
diff --git a/src/Corrade/Containers/StridedArrayView.h b/src/Corrade/Containers/StridedArrayView.h
index 24fe8a24b..930ff2e32 100644
--- a/src/Corrade/Containers/StridedArrayView.h
+++ b/src/Corrade/Containers/StridedArrayView.h
@@ -2221,7 +2221,8 @@ namespace Implementation {
     };
     template<class T> struct StridedElement<1, T> {
         static T& get(typename std::conditional<std::is_const<T>::value, const void, void>::type* data, const Size1D&, const Stride1D& stride, std::size_t i) {
-            return *reinterpret_cast<T*>(static_cast<typename std::conditional<std::is_const<T>::value, const char, char>::type*>(data) + i*stride._data[0]);
+            return *reinterpret_cast<T*>(reinterpret_cast<std::uint64_t>(data) + std::uint64_t(i)*std::int64_t(stride._data[0]));
         }
     };
 }
```

Which is obviously not a solution, but seems to confirm my suspicion that an expansion to a 64-bit value happens where it shouldn't, causing this problem.

- - -

TL;DR: I unfortunately still can't see a way how to add a regression test for this -- because as long as the final data pointer stays in 32-bit (which it likely will in any isolated test case), it is correct because it wraps around. Which means the best I can do is to try to find & update all such cases, and add juicy comments for each :/ Any better idea?

While looking into this I found three more cases in addition to what you have in this PR, and all those will likely be there for `StridedBitArrayView` as well. Yes, I can just update those and cross my fingers that it doesn't happen again in the future, but I'd really like to do better than that.

Twinklebear:
Unfortunately I'm not sure on a nice repro case for this either, I also found that it was very sensitive to code changes. When I was debugging, if I added some prints to see what the bad values where for example, it'd stop crashing. I only started getting some hint at what was going on when I actually looked at the assembly and saw that weird value in the load instruction offset.

It's interesting you mention some expansion of the value to 64-bit may be happening incorrectly, in my case I'd been building for 32-bit webassembly, so `size_t` should be 32-bit as well.

Since the stride can be negative it's maybe best to just be on the safe side and cast the `size_t` values to `ptrdiff_t` for those cases.

Twinklebear:
Well actually, one valid test case would be to have something that compiles it and greps through the disassembly for that bad load instruction, I set that up for my repro case when I opened the Emscripten bug: https://github.com/Twinklebear/repro-gltf-compile-bug . But it needs a bit more than compile the code and run it + see bad output, so it doesn't fit quite so nicely in a normal test suite

mosra:
> It's interesting you mention some expansion of the value to 64-bit may be happening incorrectly

Sorry for being unclear. I'm on 32-bit as well. The problem there, as I see it, is that the calculation of the 32-bit pointer values is somehow, somewhere, done with a 64-bit temporary due to some particular optimization, and that's what causes it to not wrap around 2^32 correctly, resulting in a `i64.load` from a very wild offset.

If that calculation would stay in 32 bits, the (negative value interpreted as unsigned causing an) overflow would correctly wrap around to a 32-bit offset in bounds. Similarly, if all values are in 64-bits (either on a 64-bit system, or when casting everything to `std::[u]int64_t` like I did above), the calculation works correctly as well, again relying on the overflow wrapping around. 

In other words, such an overflow / wraparound was happening always, on all compilers, and it always worked, until this particular optimization that breaks it. And this optimization is likely only affecting/breaking WASM32,  on a WASM64 target this wouldn't happen because calculations on 64-bit values don't get expanded to any further bit widths.

My assumption for all of the above was that pointer arithmetic is equivalent to operations on unsigned 32-/64-bit values, for which overflow and wrap around is well defined. So I _feel_ it's some sort of a miscompilation / misoptimization bug in the compiler, but it's likely impossible to prove that it's not an UB happening somewhere.

- - -

I tried with `-Wstrict-overflow` and various other flags, unfortunately none of those caused the compiler to warn. So it's just the casts alone, commited as c931dbc5b46bb8b65fc71c677191dc039fdc8029. Oh, and as a drive-by fix, in mosra/magnum@bcf61d4de8486965c29685bdf76d759333164f19 I suppressed the literal warnings you were getting in the [repro-gltf-compile-bug](https://github.com/Twinklebear/repro-gltf-compile-bug) repository.

Thanks again for your effort with investigating this!

Twinklebear:
Awesome thanks for the fix!

PR #204, opened 2025-09-23
Fix shadow-uncaptured-local warnings in Format.h

asmaloney:
Triggered when using "-Wshadow-uncaptured-local" on Apple clang version 17.0.0 (clang-1700.3.19.1)

### DIFF
--- a/src/Corrade/Utility/Format.h
+++ b/src/Corrade/Utility/Format.h
@@ -315,8 +315,8 @@ struct BufferFormatter {
     /*implicit*/ constexpr BufferFormatter(): _fn{}, _value{} {}
 
     template<class T> explicit BufferFormatter(const T& value): _value{&value} {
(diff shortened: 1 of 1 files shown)


mosra:
Yeah that was just a flaky test, where on heavily emulated targets (android / emscripten) a expected-to-be-zero benchmark somehow doesn't end up being zero. The other failures are due to AppVeyor updating their VMs to a never MSVC version, which I fixed in 41741324f69dd15b236ec764863509ecf4dd2d0e.

I tried to reproduce locally with Clang 20 (while Apple Clang 17 seems to be "vanilla" Clang 19), and it doesn't produce the warning, so I suppose this was some temporary false positive on earlier versions? Nevertheless, I pushed your change as 480cd63f6e69843f52f43ac5001b04c3c8c00e1b, together with an explanatory comment to ensure I don't try to "clean up" the variable names in the future again :)

Thank you!

